PR #43, opened 2022-10-13
Feature/remove config from topic

mehab:
Removed config topic from poc. Updated readme


src/main/java/org/acme/consumer/PrimaryConsumer.java
```diff
@@ -30,13 +30,13 @@ public class PrimaryConsumer {
     ApplicationProperty applicationProperty;
```
nscuro, line 31: Huh, didn't I remove this entire class in https://github.com/mehab/DTKafkaPOC/pull/33? o.O

mehab: yes. Added back to be able to test the code independently. Would not affect the end to end integration since we would only be sending data on EventNew topic from DT and not from the rest end point. If you want I can remove it

PR #97, opened 2022-11-17
Fix missing `CONFIGPROPERTY` table in test schema

nscuro:


nscuro:
Oh, wild. The table does exist. My bad.

PR #104, opened 2022-11-22
Fix repo meta analyzer

nscuro:
The app would fail because the analyzers still use the pooled HTTP client, which rely on Alpine's `Config` for initialization.

Initialization of `Config` would fail because neither `alpine.version` nor `application.version` were on the classpath, causing `NullPointerException`s when the `Config` class was first accessed. 

Also the URL building of the Maven analyzer was broken.


japurva1502 (approved):
LGTM

PR #115, opened 2022-11-25
Feature/unit test coverage

japurva1502:
Unit test coverage


commons/src/test/java/org/acme/commonutil/DateUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.time.Month;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.util.Date;
+
+@QuarkusTest
+public class DateUtilTest {
+
+    @Test
+    public void testParseShortDate() throws Exception {
```
sahibamittal, line 36: we can remove this test and the method _parseShortDate_ (not being used).

commons/src/test/java/org/acme/commonutil/DateUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+        Assertions.assertEquals(Month.DECEMBER, localDateTime.getMonth());
+        Assertions.assertEquals(31, localDateTime.getDayOfMonth());
+        Assertions.assertEquals(2019, localDateTime.getYear());
+        Assertions.assertEquals(15, localDateTime.getHour());
+        Assertions.assertEquals(30, localDateTime.getMinute());
+        Assertions.assertEquals(12, localDateTime.getSecond());
+    }
+
+    @Test
+    public void testDiff() {
```
sahibamittal, line 57: we can remove this test and the method DateUtil.diff (not being used).

commons/src/test/java/org/acme/common/HttpClientPoolTest.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.acme.common;
+
+
+import io.quarkus.test.junit.QuarkusTest;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
```
nscuro, line 27: `@QuarkusTest` is only needed when classes are tested that require CDI beans or other resources provided by Quarkus. As this will bootstrap all sorts of things in the background, including database containers, it's better to not use it for tests that do not need it.

VinodAnandan (approved):
Yolo LGTM

PR #121, opened 2022-11-29
Build and publish multi-platform container images

nscuro:
So the images run smooth on M1/M2 Macs as well 🧑‍💻


VinodAnandan (approved):
LGTM

PR #132, opened 2022-11-30
Fix erroneous removal of jacoco plugins

nscuro:
Also remove duplicate dependency of `kafka-streams-test-utils` in repo meta analyzer POM


VinodAnandan (approved):
LGTM

Issue #135, opened 2022-11-30
Determine when all analyzers finished their analysis for a component


nscuro:
At the moment, all analyzers do their work in parallel, and potentially at a different pace.

For CI scans however, we need a way to check when all analyzers have finished their work on a given component.

This functionality was trivial in the original API server implementation, as analyzers would run sequentially.

japurva1502:
I can take this up

japurva1502:
Approach : We can maintain an object which holds the analyzer completion information in the poc code against the purl and cpe in either kafka state store or in-memory data structure and each analyzer on completion updates its completion flag in the object and check if the other enabled analyzers are completed or not. If completed we send message to DT server.
Also the vulnerabilityResult will hold all the vulnerabilities found for the component in the format <source/ID> ( as mentioned by Niklas) in the interest of not exceeding the message size and along with that the Vulnerability Result also includes the information about whether the event was for purl, cpe or swidId.

DT code, since it sends the component from api server at the first hand knows whether the component contains cpe, purl or swidTagID and uses this information to wait for the response from poc for purl, cpe and swid.

So the information on analyzers completion is sent from poc code, however the completion on various event of component based on purl/cpe/swid is tracked by DT code, this way we need not to worry about the scaling up and down the application as all the analyzers for a particular event type of a component will run on same instance of kafka poc.

japurva1502:
Approach #2: On receiving the component, the POC can send the enabled analyzers to DT and instead of maintaining completion in kafka poc in state stores/in-memory, DT can track it (everything else remains same as in above approach)

japurva1502:
@nscuro, Currently this issue only talks about the completion for a component, should we expand it to track the completion of bom (all the components of a particular bom)?
We are currently nowhere maintaining the list of component for a bom and whether all the components are returned back to api server after analysis.

Also, as we are sending all the vulnerabilities of a component in a single message, should we skip the description for time being as limiting it to certain length is tricky as in what length to consider to have a meaningful description

nscuro:
> Currently this issue only talks about the completion for a component, should we expand it to track the completion of bom (all the components of a particular bom)?

[...]

PR #156, opened 2022-12-07
Internal analyzer

japurva1502:
This PR includes following changes:
1.Internal Analyzer code from DT to poc
2. Addition of persistent entities
3. Repositories creation for queries needed for internal analyzer
4. Integration of InternalAnalyzer with VulnerabilityTopology


commons/src/main/java/org/acme/model/Vulnerability.java
```diff
@@ -54,66 +72,109 @@ public enum Source {
...
 
+    @Column(name = "DETAIL")
+    @JsonDeserialize(using = TrimmedStringDeserializer.class)
     private String detail;
 
+    @Column(name = "RECOMMENDATION")
+    @JsonDeserialize(using = TrimmedStringDeserializer.class)
     private String recommendation;
 
+    @Column(name = "\"REFERENCES\"")
```
nscuro, line 111: You shouldn't have to use quotes here when `quarkus.hibernate-orm.database.globally-quoted-identifiers=true` is set. 

See:

https://github.com/mehab/DTKafkaPOC/blob/3eab7f96312241a13990e4cffebf0c2b2bdd6509/vulnerability-analyzer/src/main/resources/application.properties#L61-L64

japurva1502: REFERENCES is sql keyword, so naming it without quotes was throwing error

vulnerability-analyzer/src/main/java/org/acme/persistence/ComponentRepository.java
```diff
@@ -0,0 +1,16 @@
...
+
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import io.quarkus.panache.common.Parameters;
+import org.acme.model.Component;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.util.UUID;
+
+@ApplicationScoped
+public class ComponentRepository implements PanacheRepository<Component> {
```
nscuro, line 11: This class should not be necessary, see comment in `InternalAnalyzer`.

vulnerability-analyzer/src/test/java/org/acme/VulnerabilityAnalyzerTopologyTest.java
```diff
@@ -107,6 +121,39 @@ void testUnknownIdentifier() {
         Assertions.assertNull(record.value.getVulnerability());
     }
 
+    @Test
+    public void testAnalyzeOssIndex() {
```
nscuro, line 125: This method isn't really testing anything right now. The to-be-returned `result` is pre-defined, the tested code is only mapping it to a `KeyValue` object. 

Instead, I think this method should test that:
* When a component is published to the `dtrack.vuln-analysis.component` topic (`inputTopic.pipeInput(uuid, component)`)
* And OSS Index is enabled (`Mockito.when(ossIndexAnalyzerMock.isEnabled()).thenReturn(true);`)
* And OSS Index returns a result (`Mockito.when(ossIndexAnalyzerMock.analyze(Mockito.any())).thenReturn(List.of(result));`)
* Then the result topic (`outputTopic`) should contain a message for corresponding to that result, with identity `OSSINDEX_ANALYZER`

Or even better, when OSS Index returns multiple results, there should be multiple messages in `outputTopic`.

This will test that the entire topology was executed.

vulnerability-analyzer/src/main/java/org/acme/analyzer/InternalAnalyzer.java
```diff
@@ -0,0 +1,290 @@
...
+        this.cache = cache;
+    }
+
+    /**
+     * Determines if the {@link InternalAnalyzer} is capable of analyzing the specified Component.
+     *
+     * @param component the Component to analyze
+     * @return true if InternalAnalysisTask should analyze, false if not
+     */
+    public boolean isCapable(final Component component) {
```
nscuro, line 83: This method is not used anywhere.

nscuro (changes requested):
Some suggestions.

Also this PR for some reason contains changes that are already in `main`, for example the repository meta analyzer stuff. Any chance we can clean that up?

nscuro (changes requested):
Just tested this and it seems like the analyzer is not producing any messages to `dtrack.vuln-analysis.vulnerability` when no vulnerabilities have been found. This makes it impossible to know whether a component has been scanned or not (which will be necessary for #135).

When no vulnerability has been found, there should be a `VulnerabilityResult` with `vulnerability` set to `null`, like it's done for the other analyzers:

https://github.com/mehab/DTKafkaPOC/blob/65654e1b83d78cbc647d49424d6e691d07bef677/vulnerability-analyzer/src/main/java/org/acme/analyzer/OssIndexAnalyzer.java#L119-L125

https://github.com/mehab/DTKafkaPOC/blob/65654e1b83d78cbc647d49424d6e691d07bef677/vulnerability-analyzer/src/main/java/org/acme/analyzer/SnykAnalyzer.java#L79-L85

nscuro (approved):

PR #200, opened 2023-01-05
Snyk: skip unsupported PURL types

nscuro:
No point in wasting resources and API rate limit quota for requests that can never succeed.


mehab (approved):
lgtm

PR #207, opened 2023-01-06
Unify metric and tag names

nscuro:
Also enrich metrics with `thread_id` and `processor` tags, and add a new metric that will allow us to track the average size of OSS Index batches.


VinodAnandan (approved):
LGTM

PR #210, opened 2023-01-06
Feature/repo meta native

mehab:
Removed the usage of unirest client and created graalvm based native binary for repo meta analyzer and notification-publisher modules


commons/src/main/java/org/acme/common/ManagedHttpClientFactory.java
```diff
@@ -65,7 +67,11 @@
 import java.util.Arrays;
 import java.util.Map;
 
-public final class ManagedHttpClientFactory {
+@ApplicationScoped
+public class ManagedHttpClientFactory {
```
nscuro, line 71: Looking at this code, I think we can drop both `ManagedHttpClientFactory` and `ManagedHttpClient`, and replace it with an `@ApplicationScoped` `HttpClient` instead. The logic of `newManagedHttpClient` can be moved to a `@Produces` method. 

It looks like `ManagedHttpClientFactory` is more or less a workaround due to the lack of an Inversion of Control container like [ArC](https://quarkus.io/guides/cdi-reference) in vanilla DT. Instances of HttpClient are thread safe by design, so there's no need to create more than one.

repository-meta-analyzer/src/main/java/org/acme/repositories/NugetMetaAnalyzer.java
```diff
@@ -49,6 +51,9 @@ public class NugetMetaAnalyzer extends AbstractMetaAnalyzer {
             new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX"),
             new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'")
     };
+
+    @Inject
+    MetaModel meta;
```
nscuro, line 56: `MetaModel` should not be injected. Instead, create new instances in `analyze`, like it was done before.

repository-meta-analyzer/src/test/java/org/acme/repositories/NugetMetaAnalyzerTest.java
```diff
@@ -30,14 +31,22 @@
...
 import static org.mockserver.model.HttpResponse.response;
 
+@QuarkusTest
 class NugetMetaAnalyzerTest {
 
+    @Inject
+    NugetMetaAnalyzer analyzer;
+
+    @Inject
+    Component component;
```
nscuro, line 48: This should not be necessary. `Component` is not supposed to be a CDI-managed bean.

repository-meta-analyzer/src/test/java/org/acme/RepositoryMetaAnalyzerTopologyTest.java
```diff
@@ -56,14 +57,29 @@ public class RepositoryMetaAnalyzerTopologyTest {
...
+    @InjectMock
     private NpmMetaAnalyzer npmMetaAnalyzerMock;
+
+    @InjectMock
     private NugetMetaAnalyzer nugetMetaAnalyzerMock;
+
+    @InjectMock
     private PypiMetaAnalyzer pypiMetaAnalyzerMock;
+
+    @InjectMock
```
nscuro, line 82: As you're using `@InjectMock` now, you can remove the mocking logic in `beforeEach`.

nscuro (changes requested):

commons/pom.xml
```diff
@@ -50,10 +50,16 @@
     </dependency>
 
     <dependency>
-      <groupId>com.konghq</groupId>
-      <artifactId>unirest-java</artifactId>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.10</version>
```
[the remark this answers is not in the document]
mehab: updated to using javax.json

commons/src/main/java/org/acme/common/ManagedHttpClient.java
```diff
@@ -17,20 +17,25 @@
...
  */
 package org.acme.common;
-
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
```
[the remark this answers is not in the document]
mehab: Since I was injecting the ManagedHttpClientFactory in the AbstractMetaAnalyzer class, the ManagedHttpClient class had to be a bean as well.. but since we have agreed to use the Arc based approach, this change would not be there anyway

commons/src/main/java/org/acme/common/ManagedHttpClientFactory.java
mehab: I have adopted this change :)

commons/src/main/resources/application.properties
```diff
@@ -0,0 +1,9 @@
+quarkus.micrometer.enabled=true
```
nscuro, line 1: Do we need this file? I figure it will be overridden by the modules importing `commons` anyway.

.github/workflows/publishNativeBinaryRepoMetaAnalyzer.yml
```diff
@@ -0,0 +1,41 @@
+name: Publish Repo Meta Analyzer Native Binary
+
+on:
+  push:
+    branches: [ "feature/repoMetaNative" ]
```
nscuro, line 5: This should be `main`.

.github/workflows/publishNativeBinaryNotification.yml
```diff
@@ -0,0 +1,41 @@
+name: Publish Notification Publisher Native Binary
+
+on:
+  push:
+    branches: [ "feature/repoMetaNative" ]
```
nscuro, line 5: This should be `main`.

commons/src/main/java/org/acme/common/HttpClientConfiguration.java
```diff
@@ -120,39 +105,41 @@ public static ManagedHttpClient newManagedHttpClient() {
...
                         .create()
                         .loadTrustMaterial(new TrustAllStrategy())
                         .build();
-                final Registry<ConnectionSocketFactory> registry = RegistryBuilder.<ConnectionSocketFactory> create()
+                final Registry<ConnectionSocketFactory> registry = RegistryBuilder.<ConnectionSocketFactory>create()
                         .register("http", PlainConnectionSocketFactory.INSTANCE)
                         .register("https", new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE))
                         .build();
                 connectionManager = new PoolingHttpClientConnectionManager(registry);
+                new PoolingHttpClientConnectionManagerMetricsBinder(connectionManager, "metaAnalyzer").bindTo(meterRegistry);
```
nscuro, line 141: This class is in `commons`, but the name in the metrics will say `metaAnalyzer`. I haven't tested it, but perhaps you can inject the `quarkus.application.name` property here, and use that instead? https://quarkus.io/guides/all-config#quarkus-core_quarkus.application.name

That way the client should always use the name of the application it's used in.

.github/workflows/publishNativeBinaryRepoMetaAnalyzer.yml
mehab: done

.github/workflows/publishNativeBinaryNotification.yml
mehab: done

commons/src/main/resources/application.properties
mehab: done

nscuro (approved):

PR #216, opened 2023-01-09
Fix load test fixture generation

nscuro:
Some etcd image versions used manifests that are incompatible with the container library that trivy uses. The `images.txt` list wasn't correctly iterated over by `generate.sh`. The `vuln-analyzer` test script has been updated to use the generated fixtures.


VinodAnandan (approved):
LGTM

PR #219, opened 2023-01-11
Tracking of scanner completion with Kafka Streams

nscuro:
This PR implements the tracking of scanner completions on the vulnerability analyzer side. 

The application now emits an additional "completion event" to the `dtrack.vuln-analysis.result` topic when all scanners capable of scanning a given component finished their work. Based on this event, downstream systems can now deduce when they don't have to expect additional events anymore.

* Instead of using component UUIDs as event keys, we now use `ScanKey`, which is a composite key containing both an arbitrary "scan token", *and* the component UUID. This makes it possible to track multiple scan executions for the same component.
* Instead of performing a general "fan-out" on component identities at the very beginning, and populating identifier-specific topics like `dtrack.vuln-analysis.component.purl`, we now generate "scan tasks" for every scanner capable of scanning a component.
* Scan tasks are then forwarded to dedicated topics, which the scanner processors subscribe to. These topics are internal Kafka Streams topics and will be created upon first startup of the application(s). Internal topics have the additional benefit of cleaning up after themselves, see https://issues.apache.org/jira/browse/KAFKA-6150. Dedicated scanner-topics are necessary in order to provide a way to control parallelism (e.g. Snyk may require more partitions than OSS Index to provide similar throughput). The partition number is configurable via `scanner.<SCANNER_NAME>.topic-partitions` property, and defaults to `3`. **It is not possible to increase or decrease the partition count once the topics are created**. Doing so requires prior deletion of the topics and / or usage of the [application reset tool](https://kafka.apache.org/20/documentation/streams/developer-guide/app-reset-tool.html).
* As the topology now makes moderate use of `KTable`s and joins, testing it using the `TopologyTestDriver` is not really practical anymore. I thus switched the topology test over to using the Quarkus Dev Services and a containerized Redpanda broker (as explored in a previous workshop, see https://github.com/mehab/DTKafkaPOC/commit/ff3c41843c7b96756d1cfc0147e29ad0547970d6). I updated `renovate.json` so the Redpanda version we use for testing stays consistent with what we have in `docker-compose.yml`.
* There are now only two "public" topics left that the vulnerability analyzer is responsible for:
  * `dtrack.vuln-analysis.component`
  * `dtrack.vuln-analysis.result`
  * All other topics are internal

image

Companion PR for the API server: https://github.com/sahibamittal/dependency-track/pull/65

### TODOs

- [x] Track completion based on component UUID **and** some sort of scan token. If the same component is submitted for scan multiple times, we need to track this separately. By using just the component UUID, this is not possible.
- [x] Periodically clean up the KTables used in this topology to prevent them from growing endlessly. See https://developer.confluent.io/tutorials/schedule-ktable-ttl/kstreams.html for example.
- [x] Send a completion event when no scanner is applicable


nscuro:
If we get to a point where we only have to report

* Vulnerability ID
* Vulnerability Source
* Scanner that identified the above

(see https://github.com/mehab/DTKafkaPOC/issues/55#issuecomment-1356839892), we can even add de-duplication logic in here: 

* Until a completion event is received, store all incoming results in a state store
  * Again, the key for the store should be component UUID and scan token
* When a completion event is received, de-duplicate Vulnerability-ID/Source pairs, and forward a single event to the API server

```json
{
  "vulnerabilities": [
    {
      "id": "CVE-123",
      "source": "NVD",
      "attributedTo": ["INTERNAL", "OSSINDEX"]
    }
  ]
}
```

VithikaS (approved):
Thanks Niklas, great MR. Should we start using some test plugins with @should annotation so it becomes easier to track what has been tested and in which test. Again just a question in general. We could do it without the plugins also. What I mean is when we look at tests, each test tells us what specifically it is testing. It is missing in some of the existing tests probably not part of this MR

nscuro:
Thanks @VithikaS! 

> Should we start using some test plugins with @should annotation so it becomes easier to track what has been tested and in which test.

I have no experience with that, but am keen to learn more about it. Can you elaborate on this, and possibly provide / link to some examples of how this would look in practice?

VithikaS (approved):

PR #246, opened 2023-01-18
Use pre-built API server container image for demo

nscuro:
Also limit max heap size of the container to 4GB, instead of allowing it to claim up to 90% of available RAM.


VinodAnandan (approved):
LGTM

japurva1502 (commented):
LGTM

PR #247, opened 2023-01-19
NVD Mirroring

sahibamittal:


mirror-service/src/main/java/org/acme/util/FileUtil.java
```diff
@@ -30,4 +32,38 @@ public static Path getTempFileLocation(String dirName, String fileSuffix) throws
...
+    /**
+     * Writes the modification time to a timestamp file
+     * @param file the file
+     * @param modificationTime the time of the last update
+     */
+    public static void writeTimeStampFile(final File file, Long modificationTime)
+    {
+        FileWriter writer = null;
+        try {
+            writer= new FileWriter(file);
```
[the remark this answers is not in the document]
sahibamittal: @sonatype-lift ignore

mirror-service/src/test/java/org/acme/nvd/NvdToCyclonedxParserTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test
+    public void testVulnerabilityRangeEmpty() throws IOException {
+
+        String jsonFile = "src/test/resources/nvd/nvd-vuln.json";
+        String jsonString = new String(Files.readAllBytes(Paths.get(jsonFile)));
+        CveItem cveItem = nvdClientConfig.objectMapper().readValue(jsonString, CveItem.class);
+        Bom result = parser.parse(cveItem);
+        Assertions.assertNotNull(result);
+
+        Assertions.assertEquals(1, result.getComponents().size());
```
VithikaS, line 28: should we use static import ? will that make it a bit cleaner ?

mirror-service/src/main/java/org/acme/client/NvdClient.java
```diff
@@ -0,0 +1,91 @@
...
+import java.time.ZoneOffset;
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Client for the NVD REST API.
+ */
+public class NvdClient extends ContextualProcessor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NvdClient.class);
```
VithikaS, line 25: Wev are not using logger in this class. Should we just remove it to clean it up ?

mirror-service/src/main/java/org/acme/client/NvdClient.java
```diff
@@ -0,0 +1,91 @@
...
+    private String lastModifiedEpochKey = "LAST_MODIFIED_EPOCH_KEY";
+    private KeyValueStore<String, Long> store;
+    private StoreBuilder<KeyValueStore<String, Long>> lastModifiedEpochStoreBuilder;
+    private String apiKey;
+
+    Collection<DefCveItem> nvdFeeds;
```
VithikaS, line 31: Do we need private level access ?

mirror-service/src/main/java/org/acme/client/NvdClient.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Client for the NVD REST API.
+ */
+public class NvdClient extends ContextualProcessor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NvdClient.class);
+    private String lastModifiedEpochKey = "LAST_MODIFIED_EPOCH_KEY";
```
VithikaS, line 26: can this be static final also ?

mirror-service/src/main/java/org/acme/nvd/NvdMirrorHandler.java
```diff
@@ -0,0 +1,45 @@
...
+        this.client = client;
+        nvdVulnerabilities = new ArrayList<>();
+    }
+
+    public List<Bom> performMirror() {
+        long start = System.currentTimeMillis();
+        LOGGER.info("Starting NVD mirroring task");
+        getNvdFeeds();
+        long end = System.currentTimeMillis();
+        LOGGER.info("NVD mirroring complete");
```
VithikaS, line 32: should we have something like this to just keep it in one line and keep it aligned with other messages         
`LOGGER.info("NVD mirroring complete. Time spent (total): {} ms", (end - start));`

mirror-service/src/main/java/org/acme/client/NvdClient.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        storeLastModifiedRequestEpoch(lastModifiedRequest);
+        return nvdFeeds;
+    }
+
+    @Override
+    public void process(Record record) {}
+
+    @Override
+    public void close() {
```
VithikaS, line 88: we can just remove is we are only calling super. Override would not be required

mirror-service/src/main/java/org/acme/nvd/NvdToCyclonedxParser.java
```diff
@@ -0,0 +1,228 @@
...
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Parser and processor of NVD data feeds.
+ */
+public final class NvdToCyclonedxParser {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NvdToCyclonedxParser.class);
+    final Bom cdxBom = new Bom();
```
VithikaS, line 40: private level access ?

mirror-service/src/test/java/org/acme/nvd/NvdMirrorHandlerTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@QuarkusTest
+public class NvdMirrorHandlerTest {
+
+    @Inject
+    Topology topology;
```
VithikaS, line 34: private level access maybe here also ?

mirror-service/src/main/resources/application.properties
```diff
@@ -10,6 +10,7 @@ quarkus.kafka-streams.application-id=dtrack-mirror-service
 quarkus.kafka-streams.application-server=localhost:8092
 quarkus.kafka-streams.topics=\
   dtrack.vulnerability.mirror.osv,\
+  dtrack.vulnerability.mirror.nvd,\
   dtrack.vulnerability
```
nscuro, line 15: When sending BOV records to Kafka, some records exceed the max size of ~1MB. We'll need to enable compression to make this work.

We can enable zstd compression with the following property:

```ini
kafka-streams.compression.type=zstd
```

However, the producer [evaluates the message size *before* compressing it](https://stackoverflow.com/a/48304851), so we'll also have to increase the `max.request.size` property to something higher. Increasing it to 2MB does it for me:

```ini
kafka-streams.max.request.size=2097152
```

scripts/create-topics.sh
```diff
@@ -47,6 +47,10 @@ done
 vuln_analysis_topics=(
   "dtrack.vuln-analysis.component"
   "dtrack.vuln-analysis.result"
+  "dtrack.vulnerability.mirror.osv"
+  "dtrack.vulnerability.mirror.nvd"
+  "dtrack.vulnerability"
+  "dtrack.vuln-analysis.info"
```
nscuro, line 53: The `dtrack.vuln-analysis.info` has been removed in `main`.

scripts/create-topics.sh
```diff
@@ -47,6 +47,10 @@ done
 vuln_analysis_topics=(
   "dtrack.vuln-analysis.component"
   "dtrack.vuln-analysis.result"
+  "dtrack.vulnerability.mirror.osv"
+  "dtrack.vulnerability.mirror.nvd"
```
nscuro, line 51: Hmmm, considering the NVD mirroring can't be split into multiple sub-tasks like OSV can, and we really only want the NVD mirroring to happen once at any given time, would it make sense to only have one partition for `dtrack.vulnerability.mirror.nvd`? 🤔 At the moment it will have 3.

mirror-service/src/main/java/org/acme/client/NvdClient.java
```diff
@@ -0,0 +1,83 @@
...
+                     String apiKey) {
+        this.apiKey = apiKey;
+        this.nvdFeeds = new ArrayList<>();
+        this.lastModifiedEpochStoreBuilder = lastModifiedEpochStoreBuilder;
+    }
+
+    @Override
+    public void init(ProcessorContext context) {
+        super.init(context);
+        this.store = context().getStateStore(this.lastModifiedEpochStoreBuilder.name());
```
nscuro, line 43: The store has to be registered in the `Topology` somehow before it can be used. At the moment this will always return `null`.

Multiple ways to achieve this:

* Use a `ProcessorSupplier` and return the `StoreBuilder` in the [`stores`](https://kafka.apache.org/33/javadoc/org/apache/kafka/streams/processor/ConnectedStoreProvider.html#stores()) method. See `SnykProcessorSupplier` for reference: https://github.com/mehab/hyades/blob/68de966acea0fa88096e5852a700b722692b9548/vulnerability-analyzer/src/main/java/org/acme/processor/snyk/SnykProcessorSupplier.java#L59-L62
* Explicitly add the store to the `Topology` using [`streamsBuilder.addStateStore`](https://kafka.apache.org/0110/documentation/streams/developer-guide#streams_processor_statestore)

Once successfully registered, you should see a topic with a name along the lines of:

```
dtrack-mirror-service-nvd-last-modified-epoch-store-changelog
```

in Redpanda Console.

nscuro (changes requested):

mirror-service/src/test/java/org/acme/nvd/NvdMirrorHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        final var cveApiMock = Mockito.mock(NvdCveApi.class);
+        Mockito.when(cveApiMock.hasNext()).thenReturn(true);
+        Mockito.when(cveApiMock.next()).thenReturn(List.of(defCveItem));
+        final BiFunction<String, Long, NvdCveApi> cveApiSupplier = (apiKey, lastModified) -> {
+            return cveApiMock;
+        };
+
+        nvdClient = new NvdClient(nvdClientConfig.lastModifiedEpochStoreBuilder(), "test-apiKey", cveApiSupplier);
+        QuarkusMock.installMockForType(nvdClient, NvdClient.class);
```
nscuro, line 77: I think the reason why this is not working right now is that `nvdClient` is not a mock. FWIW, `installMock*` only works for actually mocked objects. 

But if we mock `NvdClient`, it will be a "dumb" mock object, so it's not what we want either.

`cveApiSupplier` [as defined in `NvdClientConfig`](https://github.com/mehab/hyades/pull/247/files#diff-dbb43d1d54ba8f1eebc2ab4adaf47e48a7eb5a4b5a1cdc9f9f6a2f59e0145408R53) is a bean that is [injected into `nvdClient`](https://github.com/mehab/hyades/pull/247/files#diff-dbb43d1d54ba8f1eebc2ab4adaf47e48a7eb5a4b5a1cdc9f9f6a2f59e0145408R29).

What we want is to provide an alternative implementation of `cveApiSupplier`, without having to mock `nvdClient`. Looks like Quarkus offers a mechanism for this: https://quarkus.io/guides/getting-started-testing#cdi-alternative-mechanism

Adding this to the test class works:

```java
@Produces
@Mock // "Mock" is a bit misleading, see linked Quarkus docs for details
@ApplicationScoped
@Named("NvdCveApiSupplier")
public BiFunction<String, Long, NvdCveApi> cveApiSupplier() throws IOException {
    String jsonFile = "src/test/resources/nvd/nvd-response.json";
    String jsonString = new String(Files.readAllBytes(Paths.get(jsonFile)));
    DefCveItem defCveItem = new NvdClientConfig().objectMapper().readValue(jsonString, DefCveItem.class);

    final var cveApiMock = mock(NvdCveApi.class);
    Mockito.when(cveApiMock.hasNext()).thenReturn(true).thenReturn(false); // Return true for the first invocation, then return false. When ALWAYS returning true, the test will never terminate.
    Mockito.when(cveApiMock.next()).thenReturn(List.of(defCveItem));
    return (apiKey, lastModified) -> cveApiMock;
}
```

The test method would then look like this:

```java
@Test
void performMirrorFromTopic() throws IOException {
    inputTopic.pipeInput("event-test-id", "");
    assertThat(outputTopic.getQueueSize()).isEqualTo(1);
    assertThat(outputTopic.readRecord()).satisfies(record ->
            assertThat(record.key()).isEqualTo("NVD/CVE-1999-1341"));
}
```

One downside is that you can define the desired behavior of the `cveApiMock` instance only once per test class. You can't use `Mockito.when` etc. in individual test methods anymore. I don't quite understand how Quarkus works in this aspect, but this has been my observation so far. If you want to test different scenarios of `cveApiMock` behaving differently, you'll have to create multiple test classes, each with their own `cveApiSupplier` producer method.

sahibamittal: Really helpful, thanks Niklas! 💯

mirror-service/src/main/java/org/hyades/nvd/NvdToCyclonedxParser.java
```diff
@@ -0,0 +1,225 @@
...
+            List<Node> nodes = config.getNodes();
+            nodes.forEach(node -> {
+                List<CpeMatch> cpeMatches = node.getCpeMatch();
+                cpeMatches.forEach(cpeMatch -> {
+                    if(cpeMatch.getVulnerable()) {
+
+                        var versionRangeAffected = new Vulnerability.Affect();
+                        versionRangeAffected.setRef(getBomRef(cpeMatch.getCriteria()));
+
+                        String uniVersionRange = "vers:"+cpeMatch.getCriteria()+"/";
```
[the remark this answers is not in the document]
sahibamittal: @sonatype-lift ignoreall

mirror-service/src/main/java/org/hyades/client/NvdClient.java
```diff
@@ -0,0 +1,103 @@
...
+    public NvdClient(StoreBuilder<KeyValueStore<String, Long>> lastModifiedEpochStoreBuilder,
+                     String apiKey, BiFunction<String, Long, NvdCveApi> cveApiSupplier) {
+        this.apiKey = apiKey;
+        this.nvdFeeds = new ArrayList<>();
+        this.lastModifiedEpochStoreBuilder = lastModifiedEpochStoreBuilder;
+        this.cveApiSupplier = cveApiSupplier;
+    }
+
+    @Override
+    public void init(ProcessorContext context) {
```
nscuro, line 49: Minor nit: Please add the generic type indicators here.

```suggestion
    public void init(ProcessorContext<String, Bom> context) {
```

mirror-service/src/main/java/org/hyades/client/NvdClient.java
```diff
@@ -0,0 +1,103 @@
...
+        return 0;
+    }
+
+    void storeLastModifiedRequestEpoch(long epoch) {
+        if (this.store != null) {
+            this.store.put(LAST_MODIFIED_EPOCH_KEY, epoch);
+        }
+    }
+
+    public void update(Record record) {
```
nscuro, line 67: Minor nit: Please add the generic type indicators here.

```suggestion
    public void update(Record<String, String> record) {
```

mirror-service/src/main/java/org/hyades/nvd/NvdToCyclonedxParser.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Parser and processor of NVD data feeds.
+ */
+public final class NvdToCyclonedxParser {
+
+    private final Bom cdxBom = new Bom();
+    private final Vulnerability cdxVuln = new Vulnerability();
```
nscuro, line 38: Using fields for these variables means that instances of `NvdToCycloneDxParser` can't safely be reused. If these variables would be created in `parse` instead, `parse` could even be made `static`.

nscuro (changes requested):
Few final cosmetic nitpicks 😅

mirror-service/src/main/java/org/hyades/nvd/NvdMirrorHandler.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.kafka.streams.processor.api.Record;
+import org.hyades.client.NvdClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class NvdMirrorHandler {
```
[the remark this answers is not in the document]
sahibamittal: Right, it's not needed anymore! will remove it 👍🏼

mirror-service/src/main/java/org/hyades/nvd/NvdToCyclonedxParser.java
sahibamittal: Yes, but in that case, there are two extracted methods `getBomRef` and `parseCpe` which are setting components in the box and returning other values. These would need to be merged back inside `parse` method.

sahibamittal: Or I can return a map of bov and values from these methods to `parse`.

mirror-service/src/main/java/org/hyades/nvd/NvdProcessorSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+@ApplicationScoped
+public class NvdProcessorSupplier implements ProcessorSupplier<String, String, String, Bom> {
+
+    private final String apiKey;
+    private final BiFunction<String, Long, NvdCveApi> cveApiSupplier;
+    private final StoreBuilder<KeyValueStore<String, Long>> lastModifiedEpochStoreBuilder;
+
+    @Inject
+    public NvdProcessorSupplier(final StoreBuilder<KeyValueStore<String, Long>> lastModifiedEpochStoreBuilder,
+                                final NvdConfig config,
```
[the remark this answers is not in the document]
nscuro: @sonatype-lift ignoreall

nscuro (approved):

PR #255, opened 2023-01-24
Build and publish native executables for amd64 and arm64

nscuro:
The current build only creates native executables for `amd64`. Container images targeting `amd64` will run excruciatingly slow on `arm64` systems, e.g. all M*X* based MacBooks.  

This PR now adds cross-compilation for `arm64` into the build, and publishes multi-platform container images. It should be noted though that cross-compiling to `arm64` takes about one hour to complete.

There are separate workflows for each module. This way, it is possible to only execute a workflow when files affecting the respective module have been changed.

The workflows run upon every push to `main`, but only allow one instance of each workflow to run at any given time. When a new workflow instance is started, previous runs of the same will be canceled. 

The following images have been created using this workflow (yes I know there's a typo in `mutliarch` 😭):

* `ghcr.io/mehab/notification-publisher:native-mutliarch`
* `ghcr.io/mehab/repository-meta-analyzer:native-mutliarch`
* `ghcr.io/mehab/vulnerability-analyzer:native-mutliarch`

> **Note**
> `native-mutliarch` was just for testing. This PR targets the previously used `latest-native` tag again.


VinodAnandan (approved):
LGTM

PR #257, opened 2023-01-24
Add justifications for various architectural decisions

### LINKED ISSUE #234: Add project justification / value proposition
We should have documentation on the *why*s of this project, and what problems we're trying to solve.

This documentation should at least address the following points:

* Why this project exists (limitations of vanilla DT regarding scalability)
  * Should include rough overview of how publish-subscribe works in vanilla DT and why it is limiting
* Why Kafka was chosen as messaging solution
  * Why not RabbitMQ, Pulsar, NATS, Liftbridge
  * Any caveats that come with Kafka - we should be transparent about those
* Why Java / Quarkus was chosen
  * e.g. availability of Kafka Streams library, option of using GraalVM native image etc.
* Why we decided against microservices (for now)

I propose to put this information in a `WTF.md` file, and link to it from `README.md` for anyone who's interested.

nscuro:
Closes #234


VinodAnandan (approved):
LGTM

PR #262, opened 2023-01-25
Revert "Update quarkus.platform.version to v2.16.0.Final"

mehab:
Reverts mehab/hyades#240


nscuro:
Looks like it's building fine with 2.16.0.Final on `main`. What's the reason for this revert?

mehab:
I thought it was failing for the native binary build on github actions which we had seen a few days ago.. anyway I had approved that pr to test the kodiak check but it seems like it is not working without required checks . So this pr

PR #236, opened 2023-01-17
Workaround for buildx build failures

nscuro:
Until https://github.com/quarkusio/quarkus/issues/30424 is fixed

Signed-off-by: nscuro <nscuro@protonmail.com>

PR #264, opened 2023-01-25
Build multi-platform images for jar-based distribution

nscuro:
I missed this part when I raised #236. We currently only build these images for `amd64`.


VinodAnandan (approved):
LGTM

PR #275, opened 2023-01-27
Compose: Add frontend container; Increase default stream threads to 3

nscuro:
hyades-apiserver does not publish `bundled` container images anymore. As such, the frontend container needs to be deployed separately for the `demo` compose profile.

I also increased the default number of Kafka Streams stream threads for the hyades services to `3`. As they're running as native executables now, it's fine to have more threads, even for cases where users have a less powerful workstation.


VinodAnandan (approved):
LGTM

PR #276, opened 2023-01-27
Limit dependabot Maven ecosystem to root directory

nscuro:
All dependency versions are managed in the parent POM, there's no need to monitor all module POMs, too.


VinodAnandan (approved):
LGTM

PR #277, opened 2023-01-27
Tweak dockerfiles

nscuro:
Eclipse Temurin doesn't provide Alpine-based images for `arm64`. Chainguard images do not include `libstdc++`, which is required by RocksDB. Switching from the UBI8 base images to `liberica-openjdk-alpine` cuts the final image size from `~480mb` to `~280mb`.

Also removed `Dockerfile`s we don't use.


VinodAnandan (approved):
LGTM

PR #293, opened 2023-01-31
minor refactoring

VithikaS:


mirror-service/src/main/java/org/hyades/osv/OsvToCyclonedxParser.java
```diff
@@ -36,156 +36,164 @@
...
-    }
-
-    public Bom parse(final JSONObject object) {
+    public static Bom parse(JSONObject object) {
+        var cyclonedxBom = new Bom();
+        var severity = Vulnerability.Rating.Severity.UNKNOWN;
 
         // initial check if advisory is valid or withdrawn
         String withdrawn = object.optString("withdrawn", null);
+        if (withdrawn != null) {
```
sahibamittal, line 48: do we not need null check on the object itself too? before trying to access the object at line 47 to avoid NPE.

mirror-service/src/main/java/org/hyades/osv/OsvToCyclonedxParser.java
```diff
@@ -36,156 +36,164 @@
 public class OsvToCyclonedxParser {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OsvToCyclonedxParser.class);
-    final Bom cyclonedxBom;
-    String ecosystem;
-    Vulnerability.Rating.Severity severity = Vulnerability.Rating.Severity.UNKNOWN;
+    private static final String SEVERITY = "severity";
+    private static final String DATABASE_SPECIFIC = "database_specific";
```
sahibamittal, line 40: May be we don't need these variables for specific two json fields. Either, we create it for each field or we skip all and use them directly while accessing the object.

VithikaS: These fields are actually repeated 3 times. I can remove it but more chances of error if same string is repeated

mirror-service/src/main/java/org/hyades/osv/OsvToCyclonedxParser.java
VithikaS: Before this method is called, a new object is created 
            var json = new JSONObject(out.toString());
            Bom osvAdvisory = OsvToCyclonedxParser.parse(json);

We will not get null here I think

mirror-service/src/main/java/org/hyades/osv/OsvToCyclonedxParser.java
nscuro: Alternatively we could just use Jackson and map the JSON to POJOs instead of parsing it manually. Would reduce code complexity, and field names are only defined once in the respective POJO.

VithikaS: Actually that would be the good and right approach. I was not sure about some of the extraction and manipulations like adding strings like 'very:' and concatenating with ecosystem etc

PR #309, opened 2023-02-02
Feature/reafctor parser

VithikaS:
Main aim of MR was to remove instance level variables where they are not required. Some part of OsvToCyclonedxParser has been tidies but rest has been left due to lack of understanding of why certain custom logic has been applied.

### DIFF
--- a/mirror-service/src/main/java/org/hyades/client/OsvClientConfig.java
+++ b/mirror-service/src/main/java/org/hyades/client/OsvClientConfig.java
@@ -23,8 +23,9 @@ public class OsvClientConfig {
     @Named("osvObjectMapper")
     public ObjectMapper objectMapper() {
         ObjectMapper objectMapper = new ObjectMapper()
-                .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE);
+                .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
         return objectMapper.registerModule(new JavaTimeModule());
     }
-
 }
--- a/mirror-service/src/main/java/org/hyades/dto/OsvDto.java
+++ b/mirror-service/src/main/java/org/hyades/dto/OsvDto.java
@@ -1,0 +1,212 @@
+package org.hyades.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import org.cyclonedx.model.ExternalReference;
+import org.cyclonedx.model.OrganizationalContact;
+import org.cyclonedx.model.vulnerability.Vulnerability;
+
+import java.io.Serializable;
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import static org.hyades.commonutil.JsonUtil.jsonStringToTimestamp;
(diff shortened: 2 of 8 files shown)


mirror-service/src/main/java/org/hyades/dto/OsvDto.java
```diff
@@ -0,0 +1,212 @@
...
+    private String summary;
+    private String details;
+    private String published;
+    private String modified;
+    private String withdrawn;
+    private List<String> aliases;
+    private List<Credit> credits;
+    private List<Reference> references;
+
+    @JsonProperty(("database_specific"))
```
mehab, line 35: can this be a more indicative name? not. very clear

mirror-service/src/main/java/org/hyades/osv/OsvMirrorHandler.java
```diff
@@ -58,16 +62,15 @@ private void unzipFolder(ZipInputStream zipIn, String ecosystem) throws IOExcept
...
+            var json = new JSONObject(out.toString());
+            Bom osvAdvisory = new OsvToCyclonedxParser(this.objectMapper).parse(json);
             if (osvAdvisory != null) {
                 osvAdvisories.add(osvAdvisory);
             }
             zipEntry = zipIn.getNextEntry();
             reader = new BufferedReader(new InputStreamReader(zipIn));
         }
         reader.close();
-        LOGGER.info(osvAdvisories.size() +" advisories mirrored successfully for ecosystem: "+ ecosystem);
```
mehab, line 71: this log seems useful

mirror-service/src/main/java/org/hyades/dto/OsvDto.java
VithikaS: I can link Osv spec doc. https://ossf.github.io/osv-schema/#database_specific-field 
I thought about it but can't think of a better name. It really is database specific. Do you have another suggestion ?

mirror-service/src/main/java/org/hyades/osv/OsvMirrorHandler.java
VithikaS: It was getting logged at 2 places. I usually log info level logs only in public methods. 
This is logged already here I think           
 LOGGER.info("OSV mirroring completed for ecosystem: {} advisories: {}",  ecosystem, osvAdvisories);

mirror-service/src/main/java/org/hyades/dto/OsvDto.java
mehab: okay yeah maybe in a comment this documentation link would be useful

mirror-service/src/main/java/org/hyades/osv/OsvMirrorHandler.java
mehab: okay :)

mirror-service/src/main/java/org/hyades/client/OsvClientConfig.java
```diff
@@ -23,8 +23,9 @@ public CloseableHttpClient httpClient() {
     @Named("osvObjectMapper")
     public ObjectMapper objectMapper() {
         ObjectMapper objectMapper = new ObjectMapper()
-                .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE);
+                .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
mehab, line 27: should this not be true?

VithikaS: I think ignoring unknown keeps it resilient to failures. If the file we are parsing has fields we are unaware of, we will parse the rest of the fields instead of falling over which I thought was better.

mehab (approved):
lgtm

PR #310, opened 2023-02-02
Add smoke tests for native images

### LINKED ISSUE #192: Add integration tests for native executables
Natively compiled applications behave differently from those running on the JVM, e.g. when it comes to reflection (see #190).

In order to identify issues with native executables early, we should have integration tests that very they're working as expected.

https://quarkus.io/guides/building-native-image#testing-the-native-executable

nscuro:
This PR add basic smoke / [integration tests](https://quarkus.io/guides/getting-started-testing#quarkus-integration-test) for native images. Tests are run for every PR to `main`, as well as before pushing container images when building on `main`.

We can only test the native images for amd64 because we cannot execute the arm64 ones. Which is fun because the arm64 images have been the most problematic so far. 🙃

Closes #192


VinodAnandan (approved):
LGTM

PR #311, opened 2023-02-03
Publish Analysers results to mirror result topic

sahibamittal:
1. Convert the DT vulnerability object to CDX.
2. Publish results on mirror result topic (dtrack.vulnerability).

Issue: https://github.com/DependencyTrack/hyades/issues/220


sahibamittal:
@sonatype-lift ignore

mirror-service/src/main/java/org/hyades/MirrorServiceTopology.java
```diff
@@ -72,6 +81,18 @@ public Topology topology() {
...
                         .withName(processorNameProduce(KafkaTopic.NEW_VULNERABILITY, "nvd_vulnerability")));
 
+        // Vulnerability analyzers stream
+        // (K,V) to be consumed as (event ScanKey, ScanResult) from analyzers result topic
+        KStream<VulnerabilityScanKey, VulnerabilityScanResult> analyzerStream = streamsBuilder
+                .stream(KafkaTopic.VULN_ANALYSIS_RESULT.getName(), Consumed
+                        .with(scanKeySerde, scanResultSerde)
+                        .withName(processorNameConsume(KafkaTopic.VULN_ANALYSIS_RESULT)));
+        analyzerStream
+                .process(analyzerProcessorSupplier, Named.as("analyzers_vulnerabilities"))
```
nscuro, line 91: If the processor does not need access to state stores, or requires other kinds of dependencies, the dedicated `ProcessorSupplier` class can be replaced with a simple lambda:

```suggestion
                .process(AnalyzerProcessor::new , Named.as("analyzers_vulnerabilities"))
```

nscuro (approved):
Looks good @sahibamittal!

Only one suggestion regarding the `ProcessorSupplier`, but that's not a blocker and I'll leave it up to you if you want to change that or not.

mirror-service/src/main/java/org/hyades/MirrorServiceTopology.java
sahibamittal: Yes, good one, thanks I've pushed the change 👍🏼

nscuro (approved):

PR #312, opened 2023-02-03
Avoid repartition caused by tombstone emitters

nscuro:
By using `FixedKeyProcessor`, we can hint to Kafka Streams that the record key does not change while `TombstoneEmittingProcessor` is processing it.

In the current topology, this avoids **2** repartition operations, and reduces the number of sub-topologies from 6 to 4.

Before:

before

After:

after


mehab (approved):
lgtm

PR #331, opened 2023-02-14
Update _build-native-meta.yml

VithikaS:
There is error in current workflow so native images are not being built. Unexpected symbol: '"amd64"'. Located at position 16 within expression: matrix.arch == "amd64"  It could be an issue with using double quote around string, though I am not sure.


VinodAnandan (approved):
LGTM

Issue #244, opened 2023-01-18
Migrate remaining events to use Protobuf schemas


nscuro:
We currently maintain duplicate code for message POJOs in the API server and PoC applications, adding unnecessary overhead, and constantly putting us in danger of various inconsistencies between model classes.

Look into defining message models in a common schema, and automatically generate POJOs from those.

Possible options:

* JSON schema (without schema registry): https://github.com/joelittlejohn/jsonschema2pojo has a Maven plugin to generate POJOs at build time
* JSON schema (with schema registry): https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-json.html
* Avro (requires schema registry): https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-avro.html
* Protobuf (requires schema registry): https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-protobuf.html

For now, I think I'd prefer to *not* introduce the schema registry. Unless Redpanda is used (which [bundles a schema registry](https://redpanda.com/blog/schema-registry)), utilizing a schema registry will mean there's an additional service that needs to be taken care of.

### Progress

- [ ] Mirror
- [x] ~Metrics (#386)~
- [x] Notifications (#415)
- [x] Repo meta analysis (#411)
- [x] Vulnerability analysis (#338)

nscuro:
This should also include proper wording of the events. For example, the payloads for `dtrack.repo-meta-analysis.component` and `dtrack-vuln-analysis.component` are [commands](https://livebook.manning.com/book/event-streams-in-action/chapter-9/).

nscuro:
At the moment only the vulnerability analyzer is using Protobuf schemas.

All modules should be switched to Protobuf to make it more uniform.

Note: We will need some (custom) tooling to still be able to test things manually.

PR #337, opened 2023-02-20
Add protobuf serializers and schemas for vulnerability analysis domain

nscuro:
This PR adds an initial draft of [Protobuf](https://protobuf.dev/) schemas for use with the vulnerability analyzer, along with some serialization logic.

Custom serialization logic was needed, as the [official Protobuf serializers](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-protobuf.html) require use of a schema registry, which we decided to not use for now.

Both serialization logic and generated Java classes reside in the new `proto` module, which is intended to be used by modules within the Hyades project. Applications living outside this repo are intended to pick the `.proto` files they're interested in (e.g. `vuln-analysis_v1.proto`), and generate the corresponding code by themselves.

In later PRs, existing code in `vulnerability-analyzer` and `mirror-service` will be updated to use these schemas.

Relates to #244

### DIFF
--- a/proto/src/main/proto/vuln-analysis_v1.proto
+++ b/proto/src/main/proto/vuln-analysis_v1.proto
@@ -19,9 +19,12 @@ message Component {
   // Package URL (PURL) of the component.
   optional string purl = 3;
 
+  // Software Identification (SWID) Tag ID of the component.
+  optional string swid_tag_id = 4;
+
   // Whether the component is internal to the organization.
   // Internal components should not be looked up in external sources.
-  bool internal = 4;
+  bool internal = 5;
 }
 
 enum Scanner {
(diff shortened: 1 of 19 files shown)


proto/src/main/proto/vuln-analysis_v1.proto
```diff
@@ -0,0 +1,70 @@
...
+option java_package = "org.hyades.proto.vulnanalysis.v1";
+
+import "vuln_v1.proto";
+
+message Component {
+  // UUID of the component in the Dependency-Track database.
+  // In case of an untracked component, a random UUID may be used.
+  string uuid = 1;
+
+  // Common Platform Enumeration (CPE) of the component.
```
mehab, line 16: should there also be swid here?

nscuro: Yup, good spot!

mehab (approved):

PR #339, opened 2023-02-22
Update nvd-lib version and imports

sahibamittal:
https://github.com/jeremylong/vuln-tools/tree/main/nvd-lib 

NVD-lib has been updated with couple of class files renamed.


sahibamittal:
@sonatype-lift ignoreall

mehab (approved):
lgtm

PR #355, opened 2023-03-03
Provide empty fallback value for `API_TOPIC_PREFIX` in `create-topics.sh`

nscuro:
Prevents the script from failing when `API_TOPIC_PREFIX` is not explicitly set.


sahibamittal (dismissed):

sahibamittal:
Addresses issue - https://github.com/DependencyTrack/hyades/issues/343

commons/src/test/java/org/hyades/common/KafkaTopicTest.java
```diff
@@ -1,21 +1,17 @@
...
 @QuarkusTest
 public class KafkaTopicTest {
 
-    @Inject
-    KafkaTopicConfig kafkaTopicConfig;
-
     @Test
     void testKafkaTopicConfig() {
         assertEquals("dtrack.vulnerability.mirror.osv", KafkaTopic.MIRROR_OSV.getName());
+        System.setProperty("api.topic.prefix", "customPrefix.");
```
nscuro, line 14: The property needs to be unset again after the test, otherwise other tests will be impacted.

Another option would be to use `@TestProfile`, you can find multiple usages of it throughout the code base if you need examples.

VithikaS (approved):

PR #361, opened 2023-03-08
Add native workflow for metrics service

VithikaS:
Added native image build and test stage for metric service and modified the docker compose file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-- package-ecosystem: docker
+  - package-ecosystem: docker
   directory: /
   schedule:
     interval: daily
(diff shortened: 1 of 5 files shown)


nscuro:
This reminds me: Could you add the `quarkus.application.name` property for the metrics service, and add a `hyades-` prefix there, too? Like it's done here: https://github.com/DependencyTrack/hyades/blob/main/vulnerability-analyzer/src/main/resources/application.properties#L3

The non-native container image is currently missing this prefix 😬: https://github.com/orgs/DependencyTrack/packages?repo_name=hyades

VithikaS:
> This reminds me: Could you add the `quarkus.application.name` property for the metrics service, and add a `hyades-` prefix there, too? Like it's done here: https://github.com/DependencyTrack/hyades/blob/main/vulnerability-analyzer/src/main/resources/application.properties#L3
> 
> The non-native container image is currently missing this prefix 😬: https://github.com/orgs/DependencyTrack/packages?repo_name=hyades

Yes, adding it. I did not push a commit. IntelliJ has changed formatting while pushing commit, need to fix that too

nscuro (approved):

PR #363, opened 2023-03-08
enable compression

VithikaS:
This change will enable compression on topics in vulnerability analyser service. 

It addresses issue  https://github.com/DependencyTrack/hyades/issues/353


mehab (approved):
lgtm

PR #364, opened 2023-03-09
Set timeouts for GitHub Actions workflows

nscuro:
Network problems can sometimes cause workflows to hang, this is an attempt to fail fast if workflows take longer than usual to complete.


mehab (approved):
lgtm

PR #371, opened 2023-03-10
Prevent Quarkus datasource dev services from starting for metrics service

nscuro:
As long as a JDBC driver is on the classpath, Quarkus will launch a testcontainer for the respective RDBMs in test- and dev-mode. By excluding the JDBC driver, this behavior can be stopped.

This is also how it's currently done for the mirror service.


VinodAnandan (approved):
LGTM

PR #385, opened 2023-03-12
Reduce memory footprint of demo Compose setup

nscuro:
* Limit max heap size of API server to `2GB`
* Limit max memory size of Redpanda to `512MB`
* Use `-native` image variants of metrics- and mirror-service


VinodAnandan (approved):
LGTM

PR #396, opened 2023-03-13
Fix vulnerability project count and add dashboard

VithikaS:
Fixed https://github.com/DependencyTrack/hyades/issues/368 and added dashboard


nscuro (approved):
Elegant solution!

VithikaS:
> Elegant solution!

Thanks :)

PR #397, opened 2023-03-13
Enable compaction for metrics topics

nscuro:
Also hardcode partitions of portfolio metrics topic to `1`, as there will only ever be one key.


VinodAnandan (approved):
LGTM

PR #408, opened 2023-03-16
Avoid erroneous WARN logs during Snyk analysis

nscuro:
We currently log a warning when the same PURL appears in multiple issues of the Snyk response.


mehab (approved):
lgtm

PR #412, opened 2023-03-18
Revert ID column types from int to long again

nscuro:
We changed it to `int` due to datanucleus/datanucleus-rdbms#455, but a fix for that bug has since been released.

The reason to revert it now is that we need the Hyades services to be compatible with the DB schema generated by the API server, which currently they are not. It works currently because we initialize Postgres with a customized schema.


VinodAnandan (approved):
LGTM

PR #417, opened 2023-03-20
Feature/fix internal analyzer

mehab:
This PR fixes issue with internal analyzer. Tested manually with purls with and without namespace and works. 
Screenshot 2023-03-20 at 14 46 38
Screenshot 2023-03-20 at 17 18 21


commons/src/main/java/org/hyades/model/Vulnerability.java
```diff
@@ -77,7 +77,7 @@ public enum Source {
 
     @Id
     @Column(name = "ID")
-    private long id;
+    private Long id;
```
nscuro, line 80: This should stay `long` to be consistent with how it's done in other classes.

vulnerability-analyzer/src/main/java/org/hyades/persistence/VulnerableSoftwareRepository.java
```diff
@@ -22,20 +22,23 @@ public List<VulnerableSoftware> getAllVulnerableSoftware(final String part, fina
...
-                    Parameters.with("purlType", purl.getType())
-                            .and("purlNamespace", purl.getNamespace())
-                            .and("purlName", purl.getName()));
-        }
-        else {
+        } else if (purl != null) {
+            if (purl.getNamespace() == null) {
+                return list("purlType = :purlType and purlName = :purlName",
+                        Parameters.with("purlType", purl.getType())
+                                .and("purlName", purl.getName()));
```
nscuro, line 34: We still need to check that the namespace is `null` here, just turns out that JPA requires special handling of `null` checks: https://stackoverflow.com/a/28562010

```suggestion
                return list("purlType = :purlType and purlNamespace is null and purlName = :purlName",
                        Parameters.with("purlType", purl.getType())
                                .and("purlName", purl.getName()));
```

nscuro (approved):

PR #422, opened 2023-03-21
Initial work for end-to-end tests

### LINKED ISSUE #391: Create automated smoke tests for API server
Currently the hyades and hyades api server consisting of multiple modules need to be tested manually after startup to see if they are functional. This needs to be automated to assist in faster deployments and testing of features. It also could be part of the build checks in github so that if the test fails after a particular feature change/ upgrade then the particular change should not be merged

nscuro:
This PRs lays the groundwork for end-to-end tests, incorporating all services and infrastructure components of the system.

Some cornerstones of the current implementation:

* [testcontainers](https://www.testcontainers.org/) is used to setup and tear down containers for every test case
* [WireMock](https://wiremock.org/) is used to verify Webhook alerts
* [GreenMail](https://greenmail-mail-test.github.io/greenmail/) is used to verify email alerts
* No Quarkus / CDI context is spun up to keep the tests simple and as lightweight as possible
* End-to-end tests live in the `e2e` module, and *must* be suffixed with `E2ET` to distinguish them from integration tests (`*IT`)
* End-to-end tests are skipped per default, enabling them requires usage of the `e2e-all` Maven profile
* End-to-end tests will be executed in CI **after** the *Publish Jar* workflow completes
  * This happens only for pushes to `main`

Closes #391

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,34 @@
+name: E2E
+
+on:
+  # As e2e tests depend on the latest container images,
+  # only trigger them after building the images completed.
+  workflow_run:
+    workflows:
+    - "Publish Jar"
+    types:
+    - completed
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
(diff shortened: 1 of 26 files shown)


VithikaS (approved):

mehab (approved):
lgtm

PR #453, opened 2023-03-29
Add filters to `instance` variable of Grafana dashboards

nscuro:
This ensures that the dashboards do not include data from ALL hyades services, but only from the services the respective dashboard is supposed to display.


VinodAnandan (approved):
LGTM

PR #458, opened 2023-03-30
Provide OSS Index credentials to E2E tests; Increase timeout

nscuro:
Should hopefully suffice to be able to run E2E tests with OSS Index in CI.

Referenced secrets have been added on the repository level in GitHub.


VinodAnandan (approved):
LGTM

PR #463, opened 2023-04-01
Enable Snappy compression for notification-publisher and repository-meta-analyzer

nscuro:
To make compression consistent across services.

Event though the notification publisher does not emit any events by itself, enabling snappy will cause the libsnappy.so library to be embedded into the native image of the service. That library is required to decompress received events.

Required for compatibility with https://github.com/DependencyTrack/hyades-apiserver/pull/79


VinodAnandan (approved):
LGTM

PR #464, opened 2023-04-02
Emit a single result event for vulnerability scans

nscuro:
We currently emit N+1 events for N capable scanners. The +1 being the completion event, which marks when no more scan results are to be expected by consumers.

While this works, it requires additional state keeping on the consumer side for various use cases (e.g. determining when no scanner reports a given vulnerability anymore, which is something we'll need to tackle soon-ish).

Also, it puts more stress on both the broker and the consumer, as more events need to be transmitted and processed. 1000 scan commands can easily morph into 4000 result events.

We initially decided against aggregating all results in a single event, out of fear that we'd hit the 1MB event size limit. Now, with both Protobuf for serialization, and Snappy for compression, even events with 80 vulnerabilities barely hit the 100kB mark. This makes aggregating results in a single event more feasible.

Dealing with only a single result event simplifies things drastically for consumers.

> **Warning**  
> This PR introduces the `dtrack.vuln-analysis.scanner.result` topic. It also modifies the names of some internal topics. I will add a list of what topic names changed to make it more transparent.


sahibamittal:
LGTM

sahibamittal (approved):

PR #472, opened 2023-04-03
Fix PURL casing mismatches for Snyk

### LINKED ISSUE #469: Result correlation and cache lookups for Snyk may fail when PURLs contain uppercase characters
Snyk responses contain PURLs in lowercase, even though the original PURL was submitted in mixed-case:

image

We do cache lookups based on "our" PURL, but populate the cache using Snyk's PURL. This will lead to erroneous cache misses.

Similarly, we correlate results returned by Snyk with the PURLs we sent, assuming they'll be equal (case sensitive). Correlation will thus fail when the original PURL contains uppercase characters.

Cache operations and correlation logic need to be adjusted so that they can deal with this.

nscuro:
Fixes #469


VinodAnandan (approved):
LGTM

PR #479, opened 2023-04-04
Cleanup

nscuro:
* Delete files that we don't need (anymore)
* Remove old topics from `KafkaTopic` enum
* Add [QueryHints](https://vladmihalcea.com/spring-read-only-transaction-hibernate-optimization/) to Hibernate queries to ensure entities are loaded in read-only mode
* Move `PublisherException` from commons to notification-publisher module, as it's not used anywhere else


sahibamittal (approved):
LGTM

PR #498, opened 2023-04-20
removing wrapper

mehab:


nscuro:
The `.mvn` directory should also be removed.

nscuro (approved):

PR #508, opened 2023-04-24
Add mapping for alias sync enabled

sahibamittal:


vulnerability-analyzer/src/main/java/org/hyades/config/InternalScannerConfig.java
```diff
@@ -9,4 +10,6 @@ public interface InternalScannerConfig {
 
     int topicPartitions();
 
+    @WithDefault("false")
+    boolean aliasSyncEnabled();
```
nscuro, line 14: The internal analyzer doesn't have any data that the API server doesn't already know, and it's only returning source and vulnId in its result. I think we can drop this config property.

vulnerability-analyzer/src/main/java/org/hyades/processor/scanner/ScanProcessorSupplier.java
```diff
@@ -20,6 +20,8 @@ public interface ScanProcessorSupplier extends ProcessorSupplier<String, ScanTas
      */
     boolean isEnabled();
 
+    boolean isAliasSyncEnabled();
```
nscuro, line 23: Do we need this? It seems like all scanners pull this value from their respective `Config` class.

VithikaS (approved):

nscuro (approved):

VinodAnandan (approved):
LGTM

PR #523, opened 2023-04-26
Remove CodeQL workflow

nscuro:
We already have enough linters and analyzers, and this one is regularly timing out on dependency downloads


VinodAnandan (approved):
LGTM

PR #546, opened 2023-05-11
Replace cpe with generic as versioning scheme in NVD parser

sahibamittal:
Issue: https://github.com/DependencyTrack/hyades/issues/486


mehab (approved):
lgtm

PR #560, opened 2023-05-17
Convert Bom to object from string

VithikaS:


proto/src/main/proto/org/hyades/notification/v1/notification.proto
nscuro, line 1: We may have to (temporarily) set
```
kafka-streams.default.deserialization.exception.handler=org.apache.kafka.streams.errors.LogAndContinueExceptionHandler
```

so that notifications in the previous format that haven't been consumed yet, or are still produced by the API server, don't crash the notification publisher.

VithikaS: Added

nscuro (approved):

nscuro:
LGTM. @VithikaS, should we make the corresponding change in the API server first before merging this?

VithikaS:
> LGTM. @VithikaS, should we make the corresponding change in the API server first before merging this?

Accompanying PR 
https://github.com/DependencyTrack/hyades-apiserver/pull/174

PR #562, opened 2023-05-18
fix SIGTERM container handling

sahibamittal:
https://github.com/DependencyTrack/hyades/issues/561 

Fix signals (e.g. SIGTERM) not being handled by the JVM process inside the container image, preventing graceful shutdown.


VinodAnandan (approved):
LGTM

PR #567, opened 2023-05-22
Temporarily disable Linux package types for Snyk

nscuro:
There are discrepancies between what SBOM generators report as PURL namespace and `distro` qualifier, and what Snyk expects.

For example, trivy uses `redhat` for both, whereas Snyk expects `rhel`. The latter is also used by Syft.

Problem is that batch requests to Snyk will fail entirely when at least one PURL in the batch does not comply with Snyk's preferred identifiers. There is no indication in the response as to which PURL caused the failure, so we can't filter them out and re-try the batch.

Disabling Linux package types for now to not impact analysis of other package types.

Reverts some changed made in #552


mehab (approved):
lgtm

PR #570, opened 2023-05-22
Helm: Add API server and frontend

nscuro:
Both are disabled per default for the time being. The API server is enabled for local Minikube deployments. To test with it, port forwarding may be used.

 I am not sure how to best setup the ingress for them in a way that does not require fiddling with `/etc/hosts` in local Minikube deployments.

Ideally, deploying to Minikube would be plug and play.


sahibamittal (approved):
lgtm

VinodAnandan (approved):
LGTM

PR #580, opened 2023-05-25
Capture physical allocated memory (RSS) of the JVM process

nscuro:
Analog to https://github.com/DependencyTrack/hyades-apiserver/pull/168


VinodAnandan (approved):
LGTM

PR #582, opened 2023-05-26
Do not use `@Transactional` in `NotificationRouter`

nscuro:
It doesn't play well with Kafka Streams.


nscuro:
Superseded by https://github.com/DependencyTrack/hyades/pull/586

PR #583, opened 2023-05-27
Revise labels in Helm chart

nscuro:
* Properly truncate `name` and `fullname` such that they never exceed 63 characters, even with suffixes
* Add `app.kubernetes.io/part-of` label
* Make `app.kubernetes.io/name` specific to individual services, instead of using a common label values across all services
* Use fully qualified names in `metadata.name`


sahibamittal (approved):
lgtm 👍🏼

PR #585, opened 2023-05-27
Replace Flyway with testcontainers init script

nscuro:
As of Quarkus 2.16.3 it is possible to provide an init script to the database testcontainer (https://github.com/quarkusio/quarkus/pull/30455).

We previously had to use Flyway to initialize the container with the API server SQL schema. This is no longer necessary.


sahibamittal (approved):
LGTM

Issue #346, opened 2023-02-27
Investigate usage of Confluent's parallel consumer


nscuro:
Confluent has published a wrapper around Kafka's Consumer/Producer API that enables parallel processing of messages, and supports [batching](https://github.com/confluentinc/parallel-consumer#batching) and [retries](https://github.com/confluentinc/parallel-consumer#retries): https://github.com/confluentinc/parallel-consumer

`parallel-consumer` can be used [alongside Kafka Streams](https://github.com/confluentinc/parallel-consumer#streams-usage-code) in the same application. It can be hooked into vert.x (which ships with Quarkus) to perform I/O operations.

Based on what I've seen so far, `parallel-consumer` may be a good fit for vulnerability analysis and repo meta analysis.

When it comes to the supported ordering guarantees, I think [key-based ordering](https://github.com/confluentinc/parallel-consumer#ordered-by-key) should suffice. These ordering guarantees are something that Quarkus' reactive messaging implementation is currently missing, see https://github.com/DependencyTrack/hyades/issues/215#issuecomment-1377347537

nscuro:
Relates to #215

PR #586, opened 2023-05-28
Migrate np to parallel consumer

nscuro:
Because it does not need any of the features provided by Kafka Streams, it doesn't necessarily make sense to use KS.

I looked into migrating to SmallRye messaging (the "native" Quarkus solution), but it still has unpleasant drawbacks that make it not a good fit (https://github.com/DependencyTrack/hyades/issues/215#issuecomment-1377347537). Most notably, ordered, blocking processing will process all events on a single thread, no matter how many partitions the input topics have. This is prone to become a huge bottleneck under high load.

With Parallel Consumer, we get ordering guarantees by key and *still* almost unlimited concurrency (https://github.com/confluentinc/parallel-consumer#ordered-by-key). Further, processing is decoupled from the actual number of partitions.

A nice side-effect is also that it supports retries (https://github.com/confluentinc/parallel-consumer#retries).

Relates to #346


sahibamittal (dismissed):
LGTM

sahibamittal:
Passing the Sonar Quality code coverage because coverage seems good locally and is optimal.

sahibamittal (approved):

PR #607, opened 2023-06-12
Lock down GitHub Actions workflows

nscuro:
Pin versions of Actions to their digest, and drop token permissions wherever possible.

Dependabot will take care of updating the actions properly, as it is capable of interpreting the `tag=vX.X.X` comment.

Token permissions are dropped globally for every GHA workflow. When specific permissions are required (e.g. to publish container images), those permissions are specified on the `jobs` level.

See also:

* https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
* https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

### DIFF
--- a/.github/workflows/_build-native-meta.yml
+++ b/.github/workflows/_build-native-meta.yml
@@ -6,6 +6,8 @@ on:
         required: true
         description: "Name of the Maven module to build"
 
+permissions: { }
+
 jobs:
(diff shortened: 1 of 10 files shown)


VinodAnandan (approved):
LGTM

PR #608, opened 2023-06-12
Fix insufficient token permissions in `_build_native_meta.yml`

nscuro:
Reusable workflows like this one cannot have more permissions than the workflow calling them (https://docs.github.com/en/actions/using-workflows/reusing-workflows#access-and-permissions). 

Because `build-native-mirror-service.yml` etc. dropped all permissions, `_build-native-meta.yml` can't request the `packages: write` permission.

Build pipelines for native images are currently failing due to this, e.g. https://github.com/DependencyTrack/hyades/actions/runs/5248293048


VinodAnandan (approved):
LGTM

PR #622, opened 2023-06-19
Fix broken native images after KStreams upgrade

nscuro:
As of kafka-streams 3.5.0, users can configure a KafkaClientSupplier, with the default implementation being `DefaultKafkaClientSupplier` (https://issues.apache.org/jira/browse/KAFKA-14395).

The supplier is called via reflection, but Quarkus <= 3.1.x doesn't yet register it as such for GraalVM.


mehab (approved):
lgtm

PR #624, opened 2023-06-19
Fix failing native image ITs not failing the build

nscuro:
The build was succeeding despite the actual integration tests failing, causing #622 to slip through.


nscuro:
Now included in #622 because it actually fixes the failing tests. There's no way to get this PR to succeed without that fix. :)

PR #625, opened 2023-06-19
renamed variable for better semantics of json result

mehab:
Addresses : https://github.com/DependencyTrack/hyades/pull/613


nscuro (approved):
Note: The `buf` failure is expected and OK. As Hyades is not yet GA we are free to introduce intentional breaking changes still.

PR #628, opened 2023-06-23
Fix native image tests for mirror-service

nscuro:
The GitHub advisory client loads template files from classpath, which are not packaged into the native image by default.


mehab (dismissed):
lgtm

VinodAnandan (approved):
LGTM

nscuro:
@VinodAnandan Thanks, but the PR was not ready yet, and the tests it was supposed to fix were still failing. It also had multiple debug statements in it that are not supposed to be in `main`. I will continue to work on this on the side, but in the meantime let's revert this merge: #629

PR #634, opened 2023-06-28
Add Benthos configs for BOM forwarding

nscuro:


scripts/benthos/bom-forward_hyades-to-hyades.yml
```diff
@@ -0,0 +1,43 @@
+---
```
mehab, line 1: this one is also going to be bom-forward_vanilla-to-hyades right?

mehab (approved):
lgtm

scripts/benthos/bom-forward_hyades-to-hyades.yml
nscuro: Not sure I understand, could you elaborate?

This file is specifically for forwarding *from* Hyades *to* another Hyades instance.

In contrast to the other file `bom-forward_vanilla-to-hyades.yml`, in this one we have to fetch the BOM from the API server, whereas vanilla DT sends the original BOM as part of the notification.

mehab: okay then my mistake. looks good :)

PR #659, opened 2023-07-10
added changes to support global authentication

mehab:
These changes are in line with the [issue](https://github.com/DependencyTrack/dependency-track/issues/2870) to make repository model uniform across hyades apiserver and hyades and also username and password are set only if authentication is needed rather than if repository in question is internal


sahibamittal (approved):
LGTM

Issue #641, opened 2023-06-29
Integration tests for native mirror-service are failing


nscuro:
The integration tests for the native version of mirror-service are broken, e.g. https://github.com/DependencyTrack/hyades/actions/runs/5408719353/jobs/9828079339?pr=638

Seems to be related to Protobuf (de-)serialization, but it works fine in the JAR version of the service.

Reproducible locally (requires GraalVM to be installed):

```shell
mvn clean package -Pnative -pl commons,commons-persistence,proto,mirror-service -DskipTests=true -Dquarkus.native.resources.includes="librocksdbjni-linux-aarch64.so,org/xerial/snappy/native/Linux/aarch64/libsnappyjava.so"
mvn -pl commons,commons-persistence,proto,mirror-service test-compile failsafe:integration-test failsafe:verify -Pnative
```

PR #676, opened 2023-07-18
Use JVM-based images for `docker-compose.yml`

nscuro:
We currently cannot guarantee that the native builds work as intended (#641). We can switch this back once we are more confident with native builds.


VinodAnandan (approved):
LGTM

PR #679, opened 2023-07-19
Remove `quarkus-maven-plugin` from `commons-persistence`

nscuro:
The plugin is only needed for applications, but not for libs. 

At the moment, building the project with `-Pnative` will cause Quarkus to build a native executable for `commons-persistence`, unnecessarily increasing build times.


VinodAnandan (approved):
LGTM

PR #705, opened 2023-07-28
Add status in ProjectVulnAnalysisCompleteSubject

sahibamittal:
https://github.com/DependencyTrack/hyades-apiserver/pull/252


nscuro:
@sahibamittal Now that native images work again, could you please add the new Proto class to the notification-publisher's `ProtoReflectionConfiguration`? 🙏

https://github.com/DependencyTrack/hyades/blob/c1e3d9bec41497063391544f8b0f3faaf10843e3/notification-publisher/src/main/java/org/hyades/notification/config/ProtoReflectionConfiguration.java#L35-L65

nscuro:
The expected notification JSON for the e2e test will also need to be updated to include the new field:

https://github.com/DependencyTrack/hyades/blob/c1e3d9bec41497063391544f8b0f3faaf10843e3/e2e/src/test/java/org/hyades/e2e/BomUploadProcessingE2ET.java#L245-L284

PR #710, opened 2023-08-01
Attempt to fix native image test failures by switching from GraalVM to Mandrel

nscuro:


nscuro:
Replaced by https://github.com/DependencyTrack/hyades/pull/711

PR #714, opened 2023-08-01
Switch to Mandrel for native image builds

nscuro:
Previous issue with installing Mandrel via GitHub Actions has been resolved, see https://github.com/graalvm/setup-graalvm/issues/54


VinodAnandan (approved):
LGTM

PR #709, opened 2023-07-31
Add release workflow

nscuro:
Releases can now be triggered via GitHub UI, under *Actions* -> *Release* -> *Run Workflow*.

Beside creating a Git tag for the release, this workflow:

* Builds and pushes JVM-based container images (no native images for now)
* Creates a GitHub release with auto generated release notes
* Generates CycloneDX BOMs for all applications and attaches them to the GitHub release

An example execution can be found here: https://github.com/nscuro/hyades/actions/runs/5718792194/job/15495340977

An example release can be found here: https://github.com/nscuro/hyades/releases/tag/v1.11.0

Closes #708

PR #729, opened 2023-08-07
Fix container image group not being lower-cased

nscuro:
Image names are not allowed to contain uppercase characters. #709 introduced a change where the `group` part of the image name would come from the GitHub repository's owner (e.g. `DependencyTrack`), which can contain uppercase characters.

Fixed by using Bash expansion on the `GITHUB_REPOSITORY_OWNER` environment variable, to convert it to lowercase.


VinodAnandan (approved):
LGTM

PR #730, opened 2023-08-07
Remove unused `org.hyades.vuln.v1` proto

nscuro:
As of https://github.com/DependencyTrack/hyades/issues/352, we use CycloneDX BOVs to communicate vulnerability information.


mehab (approved):
lgtm

PR #739, opened 2023-08-10
Fix mirror task rejection causing KStreams to fail

nscuro:
Task execution rejections and otherwise uncaught exceptions are now logged, using the logger of the respective `DatasourceMirror` class.


mehab (approved):
lgtm

PR #744, opened 2023-08-11
Run e2e tests against `snapshot` tags

nscuro:
As of https://github.com/DependencyTrack/hyades/pull/709, unreleased `main` builds are no longer tagged as `latest`, but `snapshot` instead.


mehab (approved):
lgtm

PR #781, opened 2023-09-04
Fix grammatical number of `vulnerabilities` in `ProjectVulnAnalysisCompleteSubject`

nscuro:
The field contains multiple vulnerabilities, yet its name says `vulnerability`, which is confusing.

Companion PR to https://github.com/DependencyTrack/hyades-apiserver/pull/294


mehab (approved):
lgtm

PR #788, opened 2023-09-06
Fix batches not being submitted even though `batchStore` contains records

nscuro:
When RocksDB state stores are enabled (`STATE_STORE_TYPE=rocks_db`), there was a possibility that under low-traffic circumstances, records in a scanner's `batchStore` were not submitted for analysis.

This is because the `approximateNumEntries` method can differ vastly from the *actual* number of entries in the store.

We use `approximateNumEntries` to short-circuit the punctuator responsible for submitting incomplete batches on a schedule

I temporarily modified the respective code to log both the result of `approximateNumEntries`, as well as the *actual* number of entries:

```java
final long batchStoreSize = batchStore.approximateNumEntries();
if (batchStoreSize == 0) {
    final long actual = Iterators.size(batchStore.all());
    logger.info("Current batch is empty (rocks={}, actual={})", batchStoreSize, actual);
    return;
}
```

On multiple occasions, I saw logs like the following:
```
Current batch is empty (rocks=0, actual=75)
```

In the worst case, this could cause batch entries to be "stuck" until the next wave of records comes in (which can take several hours, depending on BOM upload volume).

### DIFF
--- a/vulnerability-analyzer/src/main/java/org/hyades/processor/misc/ScannerResultAggregator.java
+++ b/vulnerability-analyzer/src/main/java/org/hyades/processor/misc/ScannerResultAggregator.java
@@ -1,6 +1,7 @@
 package org.hyades.processor.misc;
 
-import com.google.protobuf.Timestamp;
+import com.google.protobuf.util.Timestamps;
+import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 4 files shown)


mehab (approved):
lgtm

PR #795, opened 2023-09-08
Ensure consistent formatting of notification timestamps

nscuro:
Adds a custom method for formatting Protobuf `Timestamp`s.

With the default `Timestamps#toString` method, the format could change slightly, depending on how precise the provided `Timestamp` was. The number of fractional digit could either be of length 0, 3, 6, or 9. To prevent any complications in timestamp parsing, we settle for a fixed length of 3.


mehab (approved):
lgtm

mehab (approved):

PR #812, opened 2023-09-15
Feature/support go with snyk and ossIndex

mehab:
Made supported package url types configurable
Also added support for go package urls in snyk


vulnerability-analyzer/src/main/java/org/hyades/config/SnykConfig.java
```diff
@@ -19,6 +19,8 @@ public interface SnykConfig {
 
     List<SeveritySource> severitySourcePriority();
 
+    List<String> supportedPackageUrl();
```
nscuro, line 22: May I suggest using `supportedPurlTypes` instead? As this is a list the variable name should be plural. And because it is about the type part of the PURL, not the PURL entirely, it should say "types" to avoid any ambiguity.

vulnerability-analyzer/src/main/resources/application.properties
```diff
@@ -83,6 +75,7 @@ scanner.ossindex.retry.multiplier=2
 scanner.ossindex.retry.max-attempts=6
 scanner.ossindex.retry.max-duration=2M
 scanner.ossindex.alias-sync-enabled=false
+scanner.ossindex.supported-package-url=cargo,composer,gem,hex,maven,npm,nuget,pypi,rpm,conan,conda,swift,cocoapods,cran
```
nscuro, line 78: Please also add the new properties to [`CONFIGURATION.md`](https://github.com/DependencyTrack/hyades/blob/main/CONFIGURATION.md) :)

vulnerability-analyzer/src/main/java/org/hyades/config/SnykConfig.java
mehab: yes I changed this myself. I also thought it is not good enough name 👍

sahibamittal (approved):
LGTM

PR #843, opened 2023-10-09
fixed failing unit test in main

mehab:
Fixed failing unit test failing build of main branch


repository-meta-analyzer/src/test/java/org/hyades/repositories/NugetMetaAnalyzerTest.java
```diff
@@ -110,7 +110,7 @@ void testAnalyzer() throws Exception {
         Assertions.assertEquals(RepositoryType.NUGET, analyzer.supportedRepositoryType());
         Assertions.assertNotNull(metaModel.getComponent());
         Assertions.assertNotNull(metaModel.getLatestVersion());
-        Assertions.assertNotNull(metaModel.getPublishedTimestamp());
+        Assertions.assertNull(metaModel.getPublishedTimestamp());
```
VithikaS, line 113: Is publishedTimestamp the timestamp of latest version ? In that case it should be Not Null

mehab: the url that the test is going to is set to return 404 in which case the latest version of the component is the existing version that was set while sending the component in. And the date is not set so it is coming as null

VithikaS (approved):

repository-meta-analyzer/src/test/java/org/hyades/repositories/NugetMetaAnalyzerTest.java
```diff
@@ -34,6 +35,9 @@
 
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.sql.Date;
```
VithikaS, line 38: umm.. is sql date required ?

VithikaS (approved):

PR #861, opened 2023-10-19
Fix false negatives in NVD CPE matching

nscuro:
Ported from https://github.com/DependencyTrack/dependency-track/pull/3070

### DIFF
--- a/commons-persistence/pom.xml
+++ b/commons-persistence/pom.xml
@@ -36,6 +36,7 @@
             <artifactId>quarkus-jacoco</artifactId>
             <scope>test</scope>
         </dependency>
+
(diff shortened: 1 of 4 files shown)


nscuro:
Sonar is complaining about missing coverage in `VulnerableSoftwareRepository`. That class is thoroughly exercised in the new CPE matching tests, however those live in a separate module so Sonar doesn't include it in the coverage metrics correctly.

I'd prefer not to write duplicate test code just to make this metric go up, but can do it anyway in case other team members wanted it.

sahibamittal (approved):

mehab (approved):

PR #868, opened 2023-10-21
Build and publish docs upon release

nscuro:
This makes the docs' versioning functionality work properly. Users will be able to switch between doc versions.


VinodAnandan (approved):
LGTM

PR #870, opened 2023-10-23
Include Cloud SQL database connector for PostgreSQL

nscuro:
Ported from https://github.com/DependencyTrack/dependency-track/pull/2651 and https://github.com/DependencyTrack/dependency-track/pull/3102


VinodAnandan (approved):
LGTM

PR #877, opened 2023-10-26
Increase timeout of test workflows

nscuro:
As the test suites grow they sometimes take longer than 25min to run, causing unnecessary CI failures. Bumping the timeouts by 5min.


VinodAnandan (approved):
LGTM

PR #888, opened 2023-10-31
Refactored repository url usage for integrity and repo meta

sahibamittal:
Repo meta and integrity meta fetch can use different repositories based on their availabilities and data they provide.


repository-meta-analyzer/src/main/java/org/dependencytrack/processor/MetaAnalyzerProcessor.java
```diff
@@ -67,95 +67,37 @@ public void process(final FixedKeyRecord<PackageURL, AnalysisCommand> record) {
...
+                return;
             }
-            // forward result for both latest version and integrity meta
-            context().forward(record
-                    .withValue(resultBuilder.build())
-                    .withTimestamp(context().currentSystemTimeMs()));
-            return;
         }
-        // Produce "empty" result in case no repository did yield a satisfactory result.
+        // forward result with component even when no satisfactory results were yielded.
```
mehab, line 98: here actually the result trickling down also seems to be the one that has both integrity meta and repo meta information: FETCH_META_INTEGRITY_DATA_AND_LATEST_VERSION this case. So maybe update this comment

mehab (approved):
lgtm

PR #897, opened 2023-11-06
Align Protobuf version with API server

nscuro:
Instead of relying on Quarkus bringing it in transitively, specify the Protobuf version explicitly, and also use that for `protoc`.

This is analog to how it's done in the API server.


VinodAnandan (approved):
LGTM

PR #902, opened 2023-11-07
Reduce default HTTP client timeouts

nscuro:
The previous defaults of 30s for connection-, and 30s for socket timeouts are too excessive for a system that needs to maintain high-ish throughput.

These values are aligned with what's configured for the Quarkus REST client in the vulnerability analyzer.


mehab (approved):
lgtm

PR #913, opened 2023-11-16
Fix line break issues when cloning repository on Windows

nscuro:
When checking out the repository, bash scripts would use the platform's native line breaks, which when done on Windows would break the scripts for Linux containers. One of them being the `create-topics.sh` script we mount into the `redpanda-init` container.


VinodAnandan (approved):
LGTM

PR #916, opened 2023-11-16
Use custom frontend with CEL and integrity analysis support

nscuro:
We haven't officially forked the frontend yet, and the official frontend doesn't support CEL and integrity analysis yet.

This PR adds a customized v4.9.1 image that includes both functionalities. It was built from https://github.com/sahibamittal/dependency-track-frontend-upstream/commit/33d2b0086241b98c8f4a288caa938fb39362417e.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -131,7 +131,7 @@ services:
     restart: unless-stopped
 
   frontend:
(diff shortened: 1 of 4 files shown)


VinodAnandan (approved):
LGTM

PR #954, opened 2023-12-07
add nginx to docker compose

mehab:


docker-compose.yml
```diff
@@ -315,6 +315,16 @@ services:
       - monitoring
     restart: unless-stopped
 
+  nginx:
+    image: nginx:1.25.2-alpine
+    ports:
+      - "80:80"
+    volumes:
+      - "../../bundles:/usr/share/nginx/html/bundles:ro"
```
nscuro, line 323: This refers to a location outside of this repo that does not exist unless manually created. I think it would be good to provide an example bundle in the repo as well, and then refer to *that* location.

VinodAnandan:
@nscuro @mehab Is it good to include an example bundle?

nscuro:
Yeah we should have an example bundle.

VinodAnandan (approved):
LGTM

PR #1026, opened 2024-01-22
Improve logging for notification publishing

sahibamittal:
It's hard to debug missing notifications, because Dependency-Track only logs something when an error occurred during notification publishing. Further, error logs currently do not contain any information about the notification or alert rule involved, making it harder to pinpoint where something is broken.
This PR adds more information to log messages emitted during publishing. 

https://github.com/DependencyTrack/hyades/issues/983


notification-publisher/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
```diff
@@ -41,45 +43,66 @@ public abstract class AbstractWebhookPublisher implements Publisher {
...
+                    logger.debug("Response headers: %s".formatted((Object[]) response.getAllHeaders()));
+                    logger.debug("Response body: %s".formatted(EntityUtils.toString(response.getEntity())));
+                }
+            } else if (ctx.shouldLogSuccess()) {
+                logger.info("Destination acknowledged reception of notification with status code %d (%s)"
+                        .formatted(statusCode, ctx));
             } else {
                 logger.info("Destination {} acknowledged reception of notification with status code {} ({})",
                         destination, response.getStatusLine().getStatusCode(), ctx);
             }
```
nscuro, line 98: I think this `else` block is redundant now and can be removed.

notification-publisher/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
```diff
@@ -41,45 +43,66 @@ public abstract class AbstractWebhookPublisher implements Publisher {
...
         }
     }
 
+    protected void handleRequestException(final PublishContext ctx, final Logger logger, final Exception e) {
+        logger.error("Failed to send notification request (%s)".formatted(ctx), e);
+    }
```
nscuro, line 104: Can we remove this method as it is no longer used?

nscuro (approved):

PR #1042, opened 2024-01-28
Fix broken native build caused by Cloud SQL socket factory

nscuro:
As per Mandrel 23.1, the native image build fails when modules depend on `com.google.cloud.sql:postgres-socket-factory`: https://github.com/DependencyTrack/hyades/actions/runs/7686438316/job/20945288653. Apparently it uses a special way to call native libraries, which is not supported.

This change only includes the dependency when explicitly *not* building in native mode.

Also added an IntelliJ run profile to build native executables using containers, removing the requirement of having Graal / Mandrel installed.


VinodAnandan (approved):
LGTM

PR #1043, opened 2024-01-28
Tweak Kafka Streams config

nscuro:
Some config values are a bit too excessive for production use, in particular for:

* `commit.interval.ms`
* `cache.max.bytes.buffering`
* `metadata.max.age.ms`

Quoting the Quarkus docs (https://quarkus.io/guides/kafka-streams):

> Once you are ready to promote your application into production, consider changing the above configuration values. While `cache.max.bytes.buffering=10240` will move your records faster through the topology, the default value of `10485760` is more throughput-friendly. Also consider increasing `metadata.max.age.ms` from `500`, which will update cluster metadata quickly, but will generate a lot of redundant requests, to a value closer to the default of `300000`. A `commit.interval.ms` of `1000` is good for exactly-once processing, but might generate excessive load for the default at-least-once processing with the default value of `30000`.


VinodAnandan (approved):
LGTM

PR #1048, opened 2024-01-30
Use macOS M1 runner to build arm64 binaries

nscuro:
See announcement: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

Note: The build on macOS still produces a Linux binary, because we're running the build in a container (`quarkus.native.container-build=true`).


VinodAnandan (approved):
LGTM

PR #1061, opened 2024-02-08
Fix false positives in CPE matching due to ambiguous vendor/product relations

nscuro:
Ports https://github.com/DependencyTrack/dependency-track/pull/3209 from DT v4.10.0

Relates to https://github.com/DependencyTrack/hyades/issues/983


sahibamittal (approved):
Lgtm 👍🏼

PR #1103, opened 2024-02-28
Replace custom vers handling with versatile

sahibamittal:
While implementing the vulnerability mirroring capability, we wrote our own vers code to deal with version ranges.
Replace the custom vers handling with [versatile](https://github.com/nscuro/versatile).

Closes https://github.com/DependencyTrack/hyades/issues/1092


mirror-service/pom.xml
```diff
@@ -152,6 +152,12 @@
             <groupId>io.github.resilience4j</groupId>
             <artifactId>resilience4j-micrometer</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>io.github.nscuro</groupId>
+            <artifactId>versatile</artifactId>
+            <version>0.6.0</version>
+        </dependency>
```
nscuro, line 160: Can we move this to the [parent `pom.xml`'s `dependencyManagement`](https://github.com/DependencyTrack/hyades/blob/006af1dbb7e7e53790e59f999a80b1130ce283e3/pom.xml#L121) so it's consistent with how other dependencies are managed? You can then omit the `<version>` node here. Also please use a [property](https://github.com/DependencyTrack/hyades/blob/006af1dbb7e7e53790e59f999a80b1130ce283e3/pom.xml#L65-L94) to define the version.

nscuro (approved):

PR #1109, opened 2024-02-29
Fix `generate-bom-testdata.sh` failing due to removed ZAP image

nscuro:
ZAP is no longer an OWASP project, so their images have been removed from the `owasp` Docker Hub organization.

Also update the script to use a Trivy container instead of a local installation. Avoids the overhead of having to install Trivy.


VinodAnandan (approved):
LGTM

PR #1119, opened 2024-03-04
removed usage of mockserver

mehab:
Closes https://github.com/DependencyTrack/hyades/issues/993

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -67,12 +67,6 @@
     </dependency>
 
     <dependency>
-      <groupId>jakarta.ws.rs</groupId>
-      <artifactId>jakarta.ws.rs-api</artifactId>
-      <version>3.1.0</version>
-    </dependency>
-
-    <dependency>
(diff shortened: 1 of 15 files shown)


commons/pom.xml
```diff
@@ -61,6 +66,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>jakarta.ws.rs</groupId>
+      <artifactId>jakarta.ws.rs-api</artifactId>
+      <version>3.1.0</version>
+    </dependency>
```
nscuro, line 73: Is this dependency only needed for `jakarta.ws.rs.core.MediaType`? If so, I think we should rather use string literals (i.e. `application/json`) instead, rather than adding this.

nscuro (commented):
Great work, thanks!

Just a minor question about the new `jakarta.ws.rs-api` dependency.

nscuro (approved):

PR #1144, opened 2024-03-19
Reduce verbosity of Maven in CI

nscuro:
By disabling reporting of download status for dependencies. These horribly pollute build logs.


VinodAnandan (approved):
LGTM

PR #1153, opened 2024-03-21
Add license header and enforce presence with Checkstyle

nscuro:
Adds license headers to all files that were previously missing it.


VinodAnandan (approved):
LGTM

PR #1156, opened 2024-03-24
Add script to generate config docs

### LINKED ISSUE #1154: Generate configuration documentation from `application.properties` files
We have lots of configuration that is applied via `application.properties` and environment variables.

We tried to document the various options [here](https://github.com/DependencyTrack/hyades/blob/6e4d91ded8e86b5e92e3cea073f5e48f647560a9/docs/reference/configuration.md). But coverage is not great. And reality is that if it is a manual effort, it's out-of-date all the time. If we want this to work long-term, we need the creation of documentation to be automated. 

Hence, I propose the following: We parse the `application.properties` files of all services, and generate Markdown docs based on that. 

Comments in `application.properties` will be used to add additional metadata, such as the type (e.g. `integer`, `string`), the category (e.g. *Database*, *Kafka*), and whether the property is required.

For example, the following snippet in `application.properties`:

```ini
# Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
# eiusmod tempor incididunt ut labore et dolore magna aliqua.
#
# @category: Category A
# @type: string
# @required
foo.bar=baz
```

Should yield the following Markdown:

```markdown
## Category A

### foo.bar

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
eiusmod tempor incididunt ut labore et dolore magna aliqua.

* **Required**
* Type: `string`
* Default: `baz`
* Environment variable: `FOO_BAR`
```

Properties that are not intended to be modified by users should be hidden from the public documentation. If a property is annotated with `@hidden`, it must not be included in the output.

Properties of type `enum` must be annotated with `@options`, which lists all possible allowed values:

```ini
# Lorem ipsum dolor sit amet...
#
# @type: enum
# @options: com, de, org
foo.bar=
```

The solution must work for both the Alpine-based API server, as well as the Quarkus-based Hyades services.

nscuro:
The script parses a given `application.properties` file and generates a Markdown document for the properties within it. A custom Pebble template can be provided.

Closes #1154


nscuro:
Limiting this PR to just include the generated docs for the API server.

Because this requires modification of all `application.properties` files, I don't want this PR to become too large.

VinodAnandan (approved):
LGTM

PR #1164, opened 2024-03-28
Reduce resource footprint of e2e tests

nscuro:
As tests are short-lived, and each test only covers limited functionality, the apps don't really have a need for much memory at all.

* Reduce max heap size of API server from 2GB to 512MB
* Reduce max heap size of other services to 256MB


VinodAnandan (approved):
LGTM

PR #1165, opened 2024-03-29
Load cluster ID from database on startup

nscuro:
The functionality works per default for all services that depend on the `commons-persistence` module. `mirror-service` at the moment does not, hence it will use a randomly generated cluster ID that is out-of-sync with the other services.

Also include the cluster ID in `User-Agent` headers of all HTTP clients.

> [!WARNING]
> Depends on https://github.com/DependencyTrack/hyades-apiserver/pull/639

Closes https://github.com/DependencyTrack/hyades/issues/925


VinodAnandan (approved):
LGTM

mehab (approved):
lgtm

PR #1168, opened 2024-04-02
Handle duplicate issues reported by Snyk

nscuro:
We're occasionally seeing responses from Snyk that contain the exact same issue multiple times. It is unclear where those are coming from, but in any case we shouldn't forward obvious duplicates to the API server.


mehab (approved):
lgtm

PR #1170, opened 2024-04-03
Fix broken e2e tests due to Quarkus RestClient requiring CDI context

nscuro:
As of Quarkus 3.9.x(?), the RestClient can no longer be used without an CDI context being present. This is causing test failures, e.g. https://github.com/DependencyTrack/hyades/actions/runs/8535995179/job/23383687429

```
java.lang.IllegalStateException: The Reactive REST Client needs to be built within the context of a Quarkus application with a valid ArC (CDI) context running.
```

This replaces the Quarkus RestClient with OpenFeign (https://github.com/OpenFeign/feign), which is largely compatible, although requiring a few customizations when it comes to encoding and decoding.

### DIFF
--- a/e2e/pom.xml
+++ b/e2e/pom.xml
@@ -35,9 +35,18 @@
 
     <dependencies>
         <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-rest-client-jackson</artifactId>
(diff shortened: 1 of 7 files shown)


VinodAnandan (approved):
LGTM

PR #1173, opened 2024-04-04
Use `/dev/urandom` instead of `openssl rand` to generate secret key

nscuro:
This is mostly for local development purposes and / or demo setups. Installing OpenSSL can be problematic for folks behind corporate proxies. Using `/dev/urandom` should be sufficient considering that's what `openssl rand` is using behind the scenes anyway.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -316,7 +316,7 @@ services:
 
   secret-init:
     image: alpine:latest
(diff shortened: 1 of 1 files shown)


VinodAnandan (approved):
LGTM

aravindparappil46:
Thank you for this! I can confirm that `docker compose --profile demo up -d --pull always` works now!

PR #1180, opened 2024-04-09
Fix mapping of CPEs to vers ranges when version is NA (`-`)

### LINKED ISSUE #1179: Mirroring of NVD fails for CPEs with version NA (`-`)
### Current Behavior

Mirroring of the NVD currently fails when the mirror service encounters CPEs where the version is NA (`-`). 

This happens because `VersUtils#versFromNvdRange` in the *versatile* library calls `VersBuilder#build` on a `VersBuilder` without any constraints, causing a `VersException` to be thrown. 

While the aforementioned exception is catched in the code, it still causes a NPE further down the line, as `versForCpeMatch` is initialized as `null`, and code execution does not stop after catching the exception:

https://github.com/DependencyTrack/hyades/blob/d78a19441a16eb0b4fdfd27e4caa139c8c271b47/mirror-service/src/main/java/org/dependencytrack/vulnmirror/datasource/nvd/NvdToCyclonedxParser.java#L202-L208

### Steps to Reproduce

1. Trigger a NVD mirror operation by publishing an event with key `NVD` to the `dtrack.vulnerability.mirror.command` topic

### Expected Behavior

Mirroring of the NVD should not fail.

### Hyades Version

0.5.0-SNAPSHOT

### Repository Type

Hyades services

### Browser

N/A

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/hyades/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/hyades/issues) for whether this defect was already reported

nscuro:
### Description

Fixes mapping of CPEs to vers ranges when version is NA (`-`).

* Move creation of Component record with CPE *before* constructing the version range. This way we won't be dropping CPE data even when version range construction fails.
* Handle cases where `versFromNvdRange` doesn't yield a range.
* Log exceptions during vers constructions as warning instead of as debug message, to make it more obvious when things are not working as expected.

### Addressed Issue

Fixes #1179

### Additional Details

N/A

### Checklist

### DIFF
--- a/mirror-service/src/main/java/org/dependencytrack/vulnmirror/datasource/nvd/NvdToCyclonedxParser.java
+++ b/mirror-service/src/main/java/org/dependencytrack/vulnmirror/datasource/nvd/NvdToCyclonedxParser.java
@@ -35,6 +35,7 @@ import io.github.jeremylong.openvulnerability.client.nvd.Reference;
 import io.github.jeremylong.openvulnerability.client.nvd.Weakness;
 import io.github.nscuro.versatile.Comparator;
 import io.github.nscuro.versatile.Vers;
+import io.github.nscuro.versatile.VersException;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.cyclonedx.proto.v1_4.Bom;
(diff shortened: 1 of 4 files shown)


sahibamittal (approved):
LGTM

PR #1204, opened 2024-04-17
Don't label docs for `SNAPSHOT` versions as `latest`

nscuro:
### Description

Currently the documentation for `0.5.0-SNAPSHOT` is aliased as `latest`, which results in users seeing docs that are not applicable to the current stable version.

With this change, the `latest` alias will only be applied when building for a non-`SNAPSHOT` version.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


sahibamittal (approved):
lgtm 👍🏼

Issue #929, opened 2023-11-23
Revisit configuration management to support modification via UI


nscuro:
Most of the configuration options for new services are currently done via `application.properties` or environment variables.

This is in contrast to DT 4.x, which used primarily the UI. The drawback to UI configuration is that a new instance cannot easily be deployed with the desired settings in place, but requires manual adjustments.

However,  there is still a requirement to have configuration options tweakable via UI. We have to find a way to allow for both models, as apparently relying on only one doesn't cut it.

nscuro:
Another challenge with dynamic configuration is that we need a mechanism to notify services about changes that are impacting them. For example, when an API key for Snyk is changed, the vulnerability analyzer needs to be notified.

If we have the services poll the database every time they need the config, we're going to DoS the database. If we add caching, we need proper cache invalidation (hard), or have to live with eventually consistent configs (bad).

Typically, systems rely on ZooKeeper or etcd for use cases like this, but we can't introduce more such heavy dependencies at this point.

It may be worth looking into PostgreSQL's `LISTEN` / `NOTIFY` support: https://www.postgresql.org/docs/current/sql-notify.html

nscuro:
As per further discussion, Postgres' `LISTEN` / `NOTIFY` mechanism is **not** a good option, as it won't work across replicas. We specifically chose to make DB interactions of services read-only to support the use of read-only replicas.

Instead, the following option was proposed:

* Introduce a new Kafka topic that notifies about changes in the `CONFIGPROPERTY` table
  * Records on that topic should *not* carry property values, but group names (e.g. `vuln-source`) and names (e.g. `github.token`)
* Services relying on configuration from the database subscribe to the topic
  * NB: Consumers can't use the same consumer group, we need all instances to get all records
    * Kafka Streams does something similar to support [`GlobalKTable`](https://docs.confluent.io/platform/current/streams/concepts.html#globalktable)s
* Upon retrieval of a change record, applications reload their configuration from the database
  * Consumers can further decide whether a reload is necessary based on the property group(s) and name(s) in the record

It is possible that there is a slight delay between instance A and instance B of a service reloading their configuration. When users update configurations in the UI, the respective change will *eventually* make it to all affected services. This *should* be fine, but we need to make sure that it is clearly documented so users know what to expect.

nscuro:
Another complication with the approach above:

If we're assuming that database read replicas are being used, we'll run into race conditions:

* User updates configuration via UI
[...]

PR #1223, opened 2024-04-22
Introduce `commons-config` Quarkus extension

nscuro:
### Description

This enables access to values in the `CONFIGPROPERTY` table through Quarkus' configuration management framework.

Because the cluster ID introduced in https://github.com/DependencyTrack/hyades/pull/1165 is just a config property, the logic around `ClusterInfo` is no longer required.

### Addressed Issue

Relates to #929

### Additional Details

N/A

### Checklist


nscuro:
It seems like test coverage is not picked up correctly for the extension code. Might be due to the fact that extension tests are run against artificial Quarkus applications and coverage can't be observed: https://quarkus.io/guides/writing-extensions#testing-extensions

nscuro:
Moving back to draft. It turns out there is even a way to plug in custom secret decryption logic: https://smallrye.io/smallrye-config/Main/config/secret-keys/

Which is *exactly* what we need for the shared secret key! Working on an integration. Once done, we can ditch the current [`SecretDecryptor`](https://github.com/DependencyTrack/hyades/blob/main/commons/src/main/java/org/dependencytrack/common/SecretDecryptor.java).

nscuro:
FYI I added the `runtime` module of Quarkus extensions to Sonar's ignore list for test coverage reporting. Extension tests are written in the `deployment` module, which of course won't yield coverage for the code in the `runtime` module.

I figured instead of writing pointless tests full of mocks, it's best to acknowledge this fact and move on. :D

VithikaS (dismissed):
Great work!
DatabaseConfigSource resources directory is camel case, is that intentional?

VithikaS (approved):

PR #1254, opened 2024-05-08
Introduce quarkus wiremock extension

sahibamittal:
### Description

We currently use a custom `QuarkusTestResource` to use `WireMock` in our tests. The resource is duplicated across all modules that need it, since we don't have a common test module to share test-only code

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1237

### Checklist

- [ ] This PR introduces changes to the database model, and I have updated the [migration changelog](https://github.com/DependencyTrack/hyades-apiserver/tree/main/src/main/resources/migration) accordingly
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/hyades/tree/main/docs) accordingly


commons/pom.xml
```diff
@@ -75,6 +75,17 @@
...
+
+    <!--    Quarkus Wiremock extensions-->
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock-test</artifactId>
+    </dependency>
```
nscuro, line 87: These should have `<scope>test</scope>`.

mirror-service/pom.xml
```diff
@@ -128,11 +128,23 @@
...
+
+        <!--    Quarkus Wiremock extensions-->
+        <dependency>
+            <groupId>io.quarkiverse.wiremock</groupId>
+            <artifactId>quarkus-wiremock</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.wiremock</groupId>
+            <artifactId>quarkus-wiremock-test</artifactId>
+        </dependency>
```
nscuro, line 140: These should have `<scope>test</scope>`.

mirror-service/src/test/java/org/dependencytrack/vulnmirror/KafkaStreamsTopologyIT.java
```diff
@@ -288,34 +292,36 @@ record -> {
...
         @Test
         void test() throws Exception {
             // Simulate the first page of CVEs, containing 2 CVEs.
-            wireMock.stubFor(get(urlPathMatching("/.*"))
+            wireMock.register(get(urlPathMatching("/.*"))
                     .willReturn(aResponse()
                             .withStatus(200)
                             .withHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
-                            .withResponseBody(Body.ofBinaryOrText(resourceToByteArray("/datasource/osv/maven.zip"), new ContentTypeHeader(MediaType.APPLICATION_OCTET_STREAM)))));
+                            .withResponseBody(Body.ofBinaryOrText(resourceToByteArray("/datasource/osv/maven.zip"), new ContentTypeHeader(MediaType.APPLICATION_JSON)))));
```
nscuro, line 324: I think the previous value was correct here since the response body is a ZIP archive (binary), not JSON.

notification-publisher/pom.xml
```diff
@@ -80,6 +80,16 @@
...
 
+    <!--    Quarkus Wiremock extensions-->
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock-test</artifactId>
+    </dependency>
```
nscuro, line 91: These should have `<scope>test</scope>`.

pom.xml
```diff
@@ -287,6 +288,18 @@
...
+        <artifactId>quarkus-wiremock</artifactId>
+        <version>${quarkus.wiremock.version}</version>
+        <scope>provided</scope>
+      </dependency>
+      <dependency>
+        <groupId>io.quarkiverse.wiremock</groupId>
+        <artifactId>quarkus-wiremock-test</artifactId>
+        <version>${quarkus.wiremock.version}</version>
+        <scope>test</scope>
+      </dependency>
```
nscuro, line 302: The `scope` should not be defined in `dependencyManagement` of the parent POM, but in the module's `dependencies` instead. That makes it more explicit, as `scope` is defined where these dependencies are actually used. It's also more consistent with how it's done for other dependencies already.

repository-meta-analyzer/pom.xml
```diff
@@ -131,6 +131,16 @@
...
 
+    <!--    Quarkus Wiremock extensions-->
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkiverse.wiremock</groupId>
+      <artifactId>quarkus-wiremock-test</artifactId>
+    </dependency>
```
nscuro, line 142: These should have `<scope>test</scope>`.

vulnerability-analyzer/pom.xml
```diff
@@ -124,11 +124,23 @@
...
+
+        <!--    Quarkus Wiremock extensions-->
+        <dependency>
+            <groupId>io.quarkiverse.wiremock</groupId>
+            <artifactId>quarkus-wiremock</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.wiremock</groupId>
+            <artifactId>quarkus-wiremock-test</artifactId>
+        </dependency>
```
nscuro, line 136: These should have `<scope>test</scope>`.

vulnerability-analyzer/src/test/java/org/dependencytrack/vulnanalyzer/client/snyk/SnykClientTest.java
```diff
@@ -150,33 +143,21 @@ void testGetIssuesUsingBatchApiWillReturn500WhenPurlInBatchExceedLimit() throws
...
-                          },
-                          "errors": [
-                            {
-                              "id": "0d7d9f37-b729-4dab-8356-1fd22520fdcc",
-                              "detail": "An unknown error was encountered while handling your request. Please contact support.",
-                              "status": "500"
-                            }
-                          ]
-                        }""");
+                .hasMessageContaining("{\"jsonapi\":{\"version\":\"1.0\"},\"errors\":[{\"id\":\"0d7d9f37-b729-4dab-8356-1fd22520fdcc\",\"detail\":\"An unknown error was encountered while handling your request. Please contact support.\",\"status\":\"500\"}]}");
```
nscuro, line 154: Any reason for this change? Makes it a lot harder to read I think.

commons/pom.xml
```diff
@@ -75,6 +75,17 @@
       <artifactId>quarkus-junit5</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <!--    Quarkus Wiremock extensions-->
```
nscuro, line 79: Nit: This comment should be omitted. The dependency's name already says what it is, and you already grouped them such that it makes it more obvious that both are related.

nscuro (changes requested):

vulnerability-analyzer/src/test/java/org/dependencytrack/vulnanalyzer/client/snyk/SnykClientTest.java
sahibamittal: Yes it is failing for exactly same Json if extended/formatted like before, I'm not sure why.

mirror-service/src/test/java/org/dependencytrack/vulnmirror/KafkaStreamsTopologyIT.java
sahibamittal: It seems Quarkus wiremock doesn't support zip compression header type. 
When I'm using `APPLICATION_OCTET_STREAM` it throws `EOFException`. 
Similar for Epss test case, the error throws `ZipException: Not in GZIP format`

nscuro (approved):

PR #1264, opened 2024-05-17
Add date format to support offset in nuget analyser

sahibamittal:
### Description

Nuget Analyzer has list of two supported SimpleDateFormat which are allowed for parsing the dates being received from the repository, based on limited test data.

Now, as seen in logs, below date format is also received from the nugget repo. This date has offset dateTime and needs to be added to the list of supported formats in order to avoid the ParseException.

Date example: 1900-01-01T00:00:00+00:00
Format needed: yyyy-MM-dd'T'HH:mm:ss

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1263

### Checklist

- [ ] This PR introduces changes to the database model, and I have updated the [migration changelog](https://github.com/DependencyTrack/hyades-apiserver/tree/main/src/main/resources/migration) accordingly
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/hyades/tree/main/docs) accordingly

### DIFF
--- a/repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/NugetMetaAnalyzer.java
+++ b/repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/NugetMetaAnalyzer.java
@@ -47,7 +47,8 @@ public class NugetMetaAnalyzer extends AbstractMetaAnalyzer {
 
     public static final DateFormat[] SUPPORTED_DATE_FORMATS = new DateFormat[]{
             new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX"),
-            new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'")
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Thanks @sahibamittal!

It seems that DT v4.x [has the same issue](https://github.com/DependencyTrack/dependency-track/blob/4a15cdfebe99336739da6400a8b7495433fc735d/src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java#L47-L50). Would you be interested in submitting a PR for that, too?

sahibamittal:
> Thanks @sahibamittal!
> 
> It seems that DT v4.x [has the same issue](https://github.com/DependencyTrack/dependency-track/blob/4a15cdfebe99336739da6400a8b7495433fc735d/src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java#L47-L50). Would you be interested in submitting a PR for that, too?

Yes, I'll raise it 👍🏼

PR #1315, opened 2024-06-08
Decouple e2e tests into separate workflow

nscuro:
### Description

Decouples e2e tests into separate workflow.

Previously, e2e tests would only run after JVM-based container images for `snapshot` were built. They could not be started manually through GitHub, or with different image versions.

Now, they run against `snapshot` every night. They can be started manually via UI, and the image versions are configurable when doing so.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


VinodAnandan (approved):
LGTM

sahibamittal (approved):

PR #1338, opened 2024-06-17
Backport : Webhook alert token and new user alerts

sahibamittal:
### Description

This PR adds two new notification types (user_created and user_deleted) and the possibility of setting a token on the webhook alert.

### Addressed Issue

Backport change https://github.com/DependencyTrack/hyades/issues/1190

### Checklist

- [ ] I have read and understand the [contributing guidelines]


notification-publisher/src/main/java/org/dependencytrack/notification/publisher/PublishContext.java
```diff
@@ -129,6 +130,9 @@ public static PublishContext fromRecord(final ConsumerRecord<String, Notificatio
         } else if (notification.getSubject().is(VexConsumedOrProcessedSubject.class)) {
             final VexConsumedOrProcessedSubject subject = notification.getSubject().unpack(VexConsumedOrProcessedSubject.class);
             notificationSubjects.put(SUBJECT_PROJECT, Project.convert(subject.getProject()));
+        } else if (notification.getSubject().is(UserPrincipalSubject.class)) {
+            final UserPrincipalSubject subject = notification.getSubject().unpack(UserPrincipalSubject.class);
+            notificationSubjects.put("User", subject);
```
nscuro, line 135: The name should be lower-case for consistency. Consider adding a `SUBJECT_USER` constant for it.

Also `notificationSubjects` is meant to appear in logs, and sadly Proto structs tend to be quite "ugly" when stringified (via `.toString()`), so it would be better to either:

* Create a `User` `record` below, like it's done for `Component`, `Project`, etc., or
* Only pass `subject.getUsername()`

proto/src/main/proto/org/dependencytrack/notification/v1/notification.proto
```diff
@@ -216,6 +218,11 @@ message ProjectVulnAnalysisCompleteSubject {
   string token = 4;
 }
 
+message UserPrincipalSubject {
+  string username = 1;
+  string email = 2;
```
nscuro, line 223: `email` is labeled as `optional` in the Proto definition of the API server PR. It should be consistent.

nscuro (changes requested):

nscuro (approved):

PR #1382, opened 2024-07-10
Update CDX schema to v1.6

sahibamittal:
### Description

We currently use the CycloneDX Protobuf schema v1.4. Update the schema to v1.6: https://github.com/CycloneDX/specification/blob/master/schema/bom-1.6.proto 

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1333

### Checklist


nscuro:
Great work @sahibamittal, especially with testing all of this. Glad we dodged a bullet here and don't need to deal with fallback serializers!

That means we can remove the Proto file for CycloneDX v1.4 now, right?

sahibamittal:
> Great work @sahibamittal, especially with testing all of this. Glad we dodged a bullet here and don't need to deal with fallback serializers!
> 
> That means we can remove the Proto file for CycloneDX v1.4 now, right?

Thanks! And yes, we should be okay with removing 1.4 proto, I'll push it 👍🏼

nscuro (approved):

PR #1411, opened 2024-07-24
Rename kafka topic prefix for apache kafka compatibility

sahibamittal:
### Description

Creation of Kafka Streams repartition topics currently fails with `Unknown topic config name: prefix`, when using Apache Kafka instead of Redpanda. 

Rename the prefix config to avoid the error.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1392

### Checklist


nscuro:
Marking as breaking change since users who currently use this property will need to update their configs.

nscuro (approved):

PR #1457, opened 2024-08-19
Add support for fetching meta data from deps.dev

n1ckl0sk0rtge:
### Description

This PR adds basic capabilities to fetch meta data from deps.dev for a given PURL. With this code, only the related source code repositories will be fetched and stored.

- [X] I have read and understand the [contributing guidelines]
- [X] This PR implements an enhancement, and I have provided tests to verify that it works as intended

### DIFF
--- a/repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/DepsDevMetaAnalyzer.java
+++ b/repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/DepsDevMetaAnalyzer.java
@@ -1,0 +1,115 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.repometaanalyzer.repositories;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Iterator;
+import java.util.Optional;
+
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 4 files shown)


nscuro:
Thanks @n1ckl0sk0rtge! I've not forgotten about this PR, I'll try to get it reviewed this weekend! Apologies for the delay.

repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/DepsDevMetaAnalyzer.java
```diff
@@ -0,0 +1,115 @@
...
+    public boolean isApplicable(Component component) {
+        return component.getPurl() != null;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public RepositoryType supportedRepositoryType() {
+        return null; // Supported values for type are cargo, golang, maven, npm, nuget and pypi. 
+    }
```
nscuro, line 64: The analyzer would still need to be registered with `RepositoryAnalyzerFactory`, based on the PURL types it supports:

https://github.com/DependencyTrack/hyades/blob/941dd0ffa8bc05d3b8d6340bc2c7d5050e56c025/repository-meta-analyzer/src/main/java/org/dependencytrack/repometaanalyzer/repositories/RepositoryAnalyzerFactory.java#L33-L44

The current model assumes at most one analyzer per PURL type though, so we'd need to adjust this to support multiple. Question then is, should one take priority over the other? Do we execute all applicable analyzers, and if so, how do we merge all results back into one?

nscuro: I'm wondering if perhaps we should switch to deps.dev entirely for all public components. The `internal` status is already provided to the repository meta analyzer as per Protobuf definition: https://github.com/DependencyTrack/hyades/blob/941dd0ffa8bc05d3b8d6340bc2c7d5050e56c025/proto/src/main/proto/org/dependencytrack/repometaanalysis/v1/repo_meta_analysis.proto#L45-L47

In that case, we would not only source the repository from deps.dev, but also:

* Latest version
* Publish timestamp of latest version
* Publish timestamp of current version
* Hashes of current version (not sure if deps.dev provides that?)

I think that would be simpler than trying to support multiple analyzers per PURL.

Does that sound reasonable? Granted that change would be a bit larger.

san-zrl: > In that case, we would not only source the repository from deps.dev, but also:
>
> - Latest version
> - Publish timestamp of latest version
> - Publish timestamp of current version
> - Hashes of current version (not sure if deps.dev provides that?)

DevDeps accepts purls with and without version tags but returns different content:
If the version **is not** part of the purl we get a json object containing information about all available versions. If the version **is** part of the purl, we get the information about the given version enriched with additional data such as SOURCE_REPO. From this we could extract

- Given version
- Publish timestamp of this version
- version's source repo (if provided)

In DT we have purls with version tags. Finding out the **latest** version of a package would be complicated given the way DevDeps purlLookup works. We would have to query DevDeps twice, once without version tag and a second time with version tag. Probably not a good idea...

PR #1498, opened 2024-09-11
Limit memory usage of RocksDB and make it more configurable

nscuro:
### Description

Limit memory usage of RocksDB and make it more configurable.

Mostly following the guide from Confluent (https://docs.confluent.io/platform/current/streams/developer-guide/memory-mgmt.html). Per default, Kafka Streams creates a RocksDB block cache of up to 50MiB for each state store, which quickly adds up. Since RocksDB memory is off-heap, keeping it contained is a challenge.

With this change, we configure a single block cache to be used for all stores in a Kafka Streams instance. We limit the maximum cache size to 128MiB per default, but provide the ability to configure that if necessary.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist

- ~This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly~
- ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~


nscuro:
`RocksDbConfigSetter` can not be reasonably tested without excessive mocking. Tested manually.

PR #1621, opened 2025-01-14
Set vulnerability as rejected for empty snyk records to suppress it in apiserver

sahibamittal:
### Description

When SNYK sends records with empty data, they need to be reported to apiserver so that such component vulnerabilities can be suppressed. 

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1620

### Checklist

### DIFF
--- a/vulnerability-analyzer/src/main/java/org/dependencytrack/vulnanalyzer/processor/scanner/snyk/SnykProcessor.java
+++ b/vulnerability-analyzer/src/main/java/org/dependencytrack/vulnanalyzer/processor/scanner/snyk/SnykProcessor.java
@@ -18,12 +18,14 @@
  */
 package org.dependencytrack.vulnanalyzer.processor.scanner.snyk;
 
+import com.google.protobuf.Timestamp;
 import io.github.resilience4j.circuitbreaker.CircuitBreaker;
 import io.github.resilience4j.core.IntervalFunction;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.quarkus.cache.Cache;
 import jakarta.ws.rs.core.MultivaluedMap;
 import org.cyclonedx.proto.v1_6.Bom;
+import org.cyclonedx.proto.v1_6.Vulnerability;
 import org.dependencytrack.common.cwe.CweResolver;
 import org.dependencytrack.proto.vulnanalysis.internal.v1beta1.ScanTask;
 import org.dependencytrack.proto.vulnanalysis.v1.ScanKey;
@@ -49,6 +51,7 @@ import org.jboss.resteasy.reactive.common.util.MultivaluedTreeMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Instant;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


vulnerability-analyzer/src/main/java/org/dependencytrack/vulnanalyzer/processor/scanner/snyk/SnykProcessor.java
```diff
@@ -217,13 +218,18 @@ public void analyzeBatch(final List<RetryableRecord<String, ScanTask>> batch) {
             LOGGER.debug("No results were returned for coordinates {} (affecting {}/{} records in this batch)",
                     unmatchedRecords.getKey(), unmatchedRecords.getValue().size(), batch.size());
             for (final RetryableRecord<String, ScanTask> record : unmatchedRecords.getValue()) {
+                // Setting vulnerability rejected timestamp for records with empty data, to suppress them in apiserver.
+                var bomToBeSuppressed = Bom.newBuilder()
+                        .addVulnerabilities(Vulnerability.newBuilder()
+                                .setRejected(Timestamp.newBuilder().setSeconds(Instant.now().getEpochSecond()).build()).build());
```
nscuro, line 224: Is this a safe assumption to make here? The comment above reads:

> [...] there is no way for us to differentiate between "no vulnerabilities" and "unable to correlate".

Also, a vulnerability record without ID and source doesn't make sense to be rejected. My understanding was that Snyk still returned the ID, but no content for it. But this makes me think they don't return it at all anymore?

vulnerability-analyzer/src/test/java/org/dependencytrack/vulnanalyzer/processor/scanner/snyk/SnykProcessorTest.java
```diff
@@ -151,7 +151,9 @@ void testNoResults() throws Exception {
         assertThat(outputRecord.getValue().getStatus()).isEqualTo(ScanStatus.SCAN_STATUS_SUCCESSFUL);
         assertThat(outputRecord.getValue().hasFailureReason()).isFalse();
         assertThat(outputRecord.getValue().getScanner()).isEqualTo(Scanner.SCANNER_SNYK);
-        assertThat(outputRecord.getValue().getBom().getVulnerabilitiesCount()).isZero();
+        assertThat(outputRecord.getValue().getBom().getVulnerabilitiesList()).satisfiesExactlyInAnyOrder(
+                vulnerability -> assertThat(vulnerability.getRejected()).isNotNull()
```
nscuro, line 155: If there are legitimately no results, the rejection status doesn't make sense.

nscuro (changes requested):

vulnerability-analyzer/src/main/java/org/dependencytrack/vulnanalyzer/processor/scanner/snyk/SnykProcessor.java
sahibamittal: No they don't send any data i.e. `"data": []`

sahibamittal:
Closing PR, no changes needed at Hyades end.

